PR #8, opened 2018-12-05
feat: use new extension APIs for better UX

sqs:
- Add a panel action "Show/hide fuzzy matches" to references/definitions panel
- Default to using symbols locally
- Add an action to switch between using symbols locally
- Coexist with precise code intelligence better
- Only use old toggle action (that disables all `langserver/*` extensions) on pre-3.0 versions of Sourcegraph

These improvements are intended to make it possible to enable basic code intel for all users on Sourcegraph.com.

fixes https://github.com/sourcegraph/sourcegraph/issues/1175

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1,45 +1,142 @@
 import * as sourcegraph from 'sourcegraph'
-import { Handler, Config } from './handler'
+import { from, Observable } from 'rxjs'
+import { first, map, distinctUntilChanged, finalize } from 'rxjs/operators'
+import { Handler, Settings, DOCUMENT_SELECTOR } from './handler'
 
-export async function activate(): Promise<void> {
+// No-op for Sourcegraph versions prior to 3.0-preview
+const DUMMY_CTX = { subscriptions: { add: (_unsubscribable: any) => void 0 } }
+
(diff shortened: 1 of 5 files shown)


src/extension.ts
```diff
@@ -1,45 +1,142 @@
...
+                }
+            })
+        )
+        .subscribe()
+}
+
+function observableOrPromiseCompat<T>(
+    result: Observable<T> | Promise<T>
+): sourcegraph.ProviderResult<T> {
+    // HACK: Earlier extension API versions did not support providers returning observables. We can detect whether
```
lguychard: We should plan to handle the issue of extension API breakage differently - IMO there should be no backwards compatibility handling in extension code, it should be handled at the registry/extension host level.

eg.:
- If the user has activated the Sourcegraph extension E, which is only compatible with extension API vN, and his browser extension packages extension API vN+1, he receives a notification that "Incompatible extension E could not be activated"
- If the user's browser extension packages extension API vN, the extension registry will provide the latest version of E that was compatible with extension API vN, even if the very latest version of E on the registry was coded against vN+1

sqs: :+1: Moved to https://github.com/sourcegraph/sourcegraph/issues/1255

PR #11, opened 2019-01-11
feat: enable defs/refs/hovers by default

sqs:
Previously, all functionality was disabled by default (except for command palette actions). This meant that users needed to explicitly run "Enable fuzzy defs/refs" and "Enable fuzzy hovers" in the command palette after enabling this extension. It is easy for users to not know this was necessary and therefore not understand how to use this extension. Now, it enables them by default.

fix https://github.com/sourcegraph/sourcegraph/issues/1779

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -49,14 +49,16 @@ export function activate(ctx: sourcegraph.ExtensionContext = DUMMY_CTX): void {
     )
 
     ctx.subscriptions.add(
-        reregisterWhenEnablementChanges('basicCodeIntel.hover', () =>
+        reregisterWhenEnablementChanges('basicCodeIntel.hover', true, () =>
             sourcegraph.languages.registerHoverProvider(DOCUMENT_SELECTOR, {
                 provideHover: (doc, pos) => {
-                    if (
-                        !sourcegraph.configuration.get<Settings>().value[
-                            'basicCodeIntel.hover'
-                        ]
-                    ) {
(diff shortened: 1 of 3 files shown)


src/extension.ts
```diff
@@ -49,14 +49,16 @@ export function activate(ctx: sourcegraph.ExtensionContext = DUMMY_CTX): void {
...
-                            'basicCodeIntel.hover'
-                        ]
-                    ) {
+                    let value = sourcegraph.configuration.get<Settings>().value[
+                        'basicCodeIntel.hover'
+                    ]
+                    if (value === undefined) {
+                        value = true
+                    }
+                    if (!value) {
```
felixfbecker: You could shorten this to `if (sourcegraph.configuration.get<Settings>().value['basicCodeIntel.hover'] !== false)`

src/handler.ts
```diff
@@ -120,9 +120,9 @@ function resultToLocation(res: Result): sourcegraph.Location {
  * @see package.json contributes.configuration section for the configuration schema.
  */
 export interface Settings {
-    ['basicCodeIntel.enabled']?: boolean
+    ['basicCodeIntel.enabled']?: boolean // default true
     ['basicCodeIntel.definition.crossRepository']?: boolean
-    ['basicCodeIntel.hover']?: boolean
+    ['basicCodeIntel.hover']?: boolean // default true
```
felixfbecker: `/** @default true */`

PR #36, opened 2019-03-04
Add dev,ci scripts, parameterize sourcegraph dep for testing

chrismwendt:
Testing out a new `./ci` script.

### DIFF
--- a/ci
+++ b/ci
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+
+set -ex
+
+yarn -s
+
+pushd package
+yarn -s
+yarn -s run build
+popd
+
+pushd template
+yarn -s
+yarn -s run prettier
+yarn -s run typecheck
+yarn -s run build
+popd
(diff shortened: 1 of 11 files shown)


ci
```diff
@@ -0,0 +1,17 @@
...
+yarn -s
+
+pushd package
+yarn -s
+yarn -s run build
+popd
+
+pushd template
+yarn -s
+yarn -s run prettier
```
felixfbecker:
```suggestion
yarn -s run prettier-check
```

dev
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -ex
+
+yarn
+env FORCE_COLOR=0 \
```
felixfbecker: Why this?

chrismwendt: `FORCE_COLOR=0` tells Parcel not to rewrite lines in the terminal, which makes for more readable output.

PR #38, opened 2019-03-04
Fix unit tests

chrismwendt:


package/src/handler.test.ts
```diff
@@ -1,103 +1,76 @@
...
+                    languageId: 'cpp',
                     text: 'token',
                 },
-                expSearchQueries: [
-                    '\\btoken\\b case:yes file:.(h|c|hpp|cpp|m|cc)$ type:symbol repo:^github.com/foo/bar$@rev',
-                    '\\btoken\\b case:yes file:.(h|c|hpp|cpp|m|cc)$ type:file',
+                definitionPatterns: ['const\\s%s\\s='],
+                expectedSearchQueries: [
+                    // current file
+                    'const\\stoken\\s= case:yes file:.(cpp)$ type:file repo:^github.com/foo/bar$@rev file:^file.cpp$',
```
felixfbecker: I think the dots in the `file:` query should be escaped with a backslash

chrismwendt: Ah, yup, and in general the string needs to be run through something like `escapeRegexChars()` (I made that up).

PR #39, opened 2019-03-04
Add a bunch of tests

chrismwendt:


package/src/handler.test.ts
```diff
@@ -74,3 +80,228 @@ describe('search requests', () => {
...
+            pass
+        `,
+                definitionLine: 1,
+                commentStyle: pythonStyle,
+            }),
+            'docstring'
+        )
+    })
+
+    it('finds multi-line python doc block', async () => {
```
lguychard: All of these tests are `async` functions, but the test body seems sync. Am I missing something?

package/src/handler.ts
```diff
@@ -329,6 +329,131 @@ export function referencesQueries({
...
+    }
+
+    function findDocstringInBlockComment({
+        block: { startRegex, lineNoiseRegex, endRegex },
+        lines,
+    }: {
+        block: BlockCommentStyle
+        lines: string[]
+    }): string[] | undefined {
+        // ⚠️ Local mutation
```
lguychard: Why is the local mutation needed?

package/src/handler.test.ts
chrismwendt: I copy-pasted, so no you're not missing anything 😄

package/src/handler.ts
chrismwendt: IMO, it turned out to be easier to understand by writing imperative steps on a mutable array of lines.

PR #61, opened 2019-03-17
Add debug mode

chrismwendt:
When you set `"codeintel.debug": true`, each line that contains a symbol definition will be highlighted:

image

### DIFF
--- a/package/src/handler.ts
+++ b/package/src/handler.ts
@@ -1,5 +1,5 @@
 import { API, Result, parseUri } from './api'
-import { takeWhile, dropWhile, sortBy, flatten } from 'lodash'
+import { takeWhile, dropWhile, sortBy, flatten, omit } from 'lodash'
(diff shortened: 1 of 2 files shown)


package/src/handler.ts
```diff
@@ -669,4 +675,48 @@ export class Handler {
...
+            if (!editor) {
+                console.log('NO EDITOR')
+            } else {
+                const { repo, rev, path } = parseUri(doc.uri)
+
+                const r = await this.api.search(
+                    `repo:${repo}@${rev} file:${path} type:symbol ^` // ^ matches everything (can't leave out a query)
+                )
+                editor.setDecorations(
+                    this.sourcegraph.app.createDecorationType(),
```
lguychard: This decoration type should be created at the top-level and reused, not generated every time `setDecorations()` is called.

PR #66, opened 2019-03-20
Avoid templating TypeScript, enables callbacks

chrismwendt:
Previously, the generator would write out a TypeScript file using string interpolation, which only worked for arguments that were `JSON.stringify`able.

Now, the arguments are referred to directly, so callbacks can be added.

### DIFF
--- a/template/src/extension.ts
+++ b/template/src/extension.ts
@@ -1,6 +1,6 @@
 import { activateBasicCodeIntel } from '../../package/lib'
 import * as sourcegraph from 'sourcegraph'
-import { languageSpecs } from './languages'
+import { languageSpecs } from '../../languages'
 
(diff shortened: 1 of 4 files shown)


template/src/extension.ts
```diff
@@ -1,9 +1,20 @@
...
-        activateBasicCodeIntel(language.handlerArgs)(ctx)
+    // This is set to an individual language ID by the generator script.
+    const languageID = 'all'
+
+    if (languageID === 'all') {
+        for (const languageSpec of languageSpecs) {
+            activateBasicCodeIntel(languageSpec.handlerArgs)(ctx)
+        }
+    } else {
+        // TODO consider Record<LanguageID, LanguageSpec>
```
felixfbecker: What's preventing this?

chrismwendt: Time

PR #98, opened 2019-04-05
Remove same-file search query

chrismwendt:
Prior to this change, basic code intel would run 3 separate searches, returning early when one returns any results:

- Current file
- Current repository
- All repositories (unless on Sourcegraph.com)

This prevented basic code intel from showing valid results and letting the user choose (see https://github.com/sourcegraph/sourcegraph/issues/3234)

After this change, basic code intel will run 2 searches (skipping the current file search):

- Current repository
- All repositories (unless on Sourcegraph.com)

One reason why I didn't do this earlier is because j2d was very imprecise. Filtering by imports vastly improves this and tips the scale.

Fixes https://github.com/sourcegraph/sourcegraph/issues/3234 it'll show a choice after this change instead of jumping to the `Main` defined in the same file:

image

### DIFF
--- a/package/src/handler.test.ts
+++ b/package/src/handler.test.ts
@@ -22,8 +22,6 @@ describe('search requests', () => {
                     text: 'token',
                 },
                 expectedSearchQueries: [
-                    // current file symbols
-                    '^token$ case:yes file:.(cpp)$ type:symbol repo:^github.com/foo/bar$@rev file:^file.cpp$',
                     // current repo symbols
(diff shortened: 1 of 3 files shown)


felixfbecker:
In Go specifically, it should probably also look at the container of symbols / the package name prefix in front of the reference

chrismwendt:
Yeah, looking at the package name prefix in front of the token (i.e. to the left of the `.`) would probably improve the precision a LOT, updated https://github.com/sourcegraph/sourcegraph/issues/2298

PR #99, opened 2019-04-18
Add Objective-C

chrismwendt:
This is a draft because the ctags for Objective-C are pretty wonky. They contain a lot of colons and parent scopes:

image

https://sourcegraph.com/github.com/react-native-community/react-native-webview/-/blob/ios/RNCWKWebView.m#L242

I don't intend to merge this in until that's been cleaned up or I figure out how to deal with it in basic-code-intel.

### DIFF
--- a/languages.ts
+++ b/languages.ts
@@ -336,6 +336,15 @@ export const languageSpecs: LanguageSpec[] = [
     },
     {
         handlerArgs: {
+            languageID: 'objective-c',
+            fileExts: ['m', 'mm', 'h'],
+            docstringIgnore: /^\s*@/,
+            commentStyle: { ...cStyle, lineRegex: /\/\/\/?\s?/ },
+        },
+        stylized: 'Objective-C',
+    },
+    {
+        handlerArgs: {
             languageID: 'rust',
             fileExts: ['rs', 'rs.in'],
             docstringIgnore: /^#/,


sqs:
@chrismwendt Those are not wonky, those are how people expect it. For exampe, here is Xcode:

image

It sounds like you think that the middle results in that screenshot should just read `[NSString stringWithContentsOfFile:]` and not distinguish between the two variants (`encoding:error:` and `usedEncoding:error:`). That is incorrect. Users do want to distinguish between them.

In fact, the symbol rows should also contain the type name (the `NSString` in the screenshot).

@nicksnyder Can you confirm (as another formerly-Objective-C-dev)?

nicksnyder:
@sqs Is correct on both counts. The symbol names should include variable labels and the type of the message receiver (e.g. `[NSString stringWithContentsOfFile:encoding:error:]`)

chrismwendt:
Ah, I didn't know what Objective-C message definitions and calls looked like until I read https://blog.teamtreehouse.com/the-beginners-guide-to-objective-c-methods Here's an example:

```objective-c
// Defines a function `addX:toY:` that takes 2 int arguments and returns an int
- (int)addX:(int)x toY:(int)y {
    int sum = x + y;
    return sum;
}

// Calls the `addX:toY:` function
int newScore = [score addX:24 toY:points];
```

I see that basic-code-intel will need to find and combine the `addX:` and `toY:` pieces to form the full name `addX:toY:` in order to do a symbol search.

The type of the message receiver currently appears after the symbol name (see screenshot). The way this gets rendered could be language-specific.

image

efritz:
Can we close this with the work @creachadair has done?

chrismwendt:
I believe this is still relevant:

> basic-code-intel will need to find and combine the addX: and toY: pieces to form the full name addX:toY: in order to do a symbol search

PR #123, opened 2019-07-19
Add toggle for global symbol search

chrismwendt:
This adds a setting to disable global symbol search:

```json
  "codeIntel.globalSymbolSearch": false
```

Defaults to **`true`** (the current behavior).

The downside is that when this is `false`, cross-repository go-to-definition will stop working.

See https://sourcegraph.slack.com/archives/C08JA8Q1H/p1563496178024000?thread_ts=1563233415.015100&cid=C08JA8Q1H

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,1 @@
+comment: off
--- a/package/src/handler.ts
+++ b/package/src/handler.ts
@@ -285,12 +285,12 @@ export function definitionQueries({
     searchToken,
     doc,
     fileExts,
-    isSourcegraphDotCom,
+    enableGlobalSymbolSearch,
 }: {
     searchToken: string
     doc: TextDocument
     fileExts: string[]
-    isSourcegraphDotCom: boolean
+    enableGlobalSymbolSearch: boolean
 }): string[] {
     const queryIn = (scope: Scope): string =>
         makeQuery({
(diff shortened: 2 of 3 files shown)


codecov.yml
```diff
@@ -0,0 +1 @@
+comment: off
```
felixfbecker: As a reviewer, I rely on these comments to see if the PR has tests. What's the reason to disable it?

package/src/handler.ts
```diff
@@ -713,13 +713,26 @@ export class Handler {
...
+        const globalSymbolSearchSetting =
+            this.sourcegraph.configuration
+                .get()
+                .get('codeIntel.globalSymbolSearch') === undefined
+                ? true
+                : Boolean(
+                      this.sourcegraph.configuration
+                          .get()
+                          .get('codeIntel.globalSymbolSearch')
+                  )
```
felixfbecker: I find this ternary very complicated - you could just write:

```ts
this.sourcegraph.configuration.get().get('codeIntel.globalSymbolSearch') !== false
```

codecov.yml
chrismwendt: I find the Codecov comments distracting. Is there a way to configure visibility by user?

package/src/handler.ts
chrismwendt: Simplified ternary

PR #124, opened 2019-07-19
Add LSIF support

chrismwendt:
This tries LSIF before falling back to basic-code-intel (symbol and text search).

Most of the code is copied from https://github.com/sourcegraph/sourcegraph/blob/598201b6678c4819bd0f63e9095809fc34de82a2/lsif/extension/src/lsif.ts Once this is merged into all of the individual language extensions, the `sourcegraph/lsif` extension in sourcegraph/sourcegraph can be deleted.


package/src/index.ts
```diff
@@ -20,25 +22,188 @@ export function activateBasicCodeIntel(
...
+    return url.hash.slice(1)
+}
+
+function setPath(doc: sourcegraph.TextDocument, path: string): string {
+    const url = new URL(doc.uri)
+    url.hash = path
+    return url.href
+}
+
+async function send({
```
lguychard: Consider giving this a more explicit name:

```suggestion
async function queryLSIF({
```

package/src/index.ts
```diff
@@ -20,25 +22,188 @@ export function activateBasicCodeIntel(
...
+    if (!sourcegraph.configuration.get().get('codeIntel.lsif')) {
+        return false
+    }
+
+    if (lsifDocs.has(doc.uri)) {
+        return lsifDocs.get(doc.uri)!
+    }
+
+    const url = new URL('.api/lsif/exists', sourcegraph.internal.sourcegraphURL)
+    url.searchParams.set('repository', repositoryFromDoc(doc))
```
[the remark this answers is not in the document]
chrismwendt: Good catch! I'll insert a GraphQL request above, which will get blocked if it's a private repository.

package/src/index.ts
```diff
@@ -8,29 +8,48 @@ export { Handler, HandlerArgs, registerFeedbackButton } from './handler'
 // No-op for Sourcegraph versions prior to 3.0-preview
 const DUMMY_CTX = { subscriptions: { add: (_unsubscribable: any) => void 0 } }
 
-export function activateBasicCodeIntel(args: HandlerArgs): (ctx: sourcegraph.ExtensionContext) => void {
-    return function activate(ctx: sourcegraph.ExtensionContext = DUMMY_CTX): void {
+export function activateBasicCodeIntel(
+    args: HandlerArgs
+): (ctx: sourcegraph.ExtensionContext) => void {
```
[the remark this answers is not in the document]
chrismwendt: Currying it made other code more ergonomic in the past, but that's no longer the case. I'm going to leave it for the time being, but it'd be nice to uncurry it.

chrismwendt:
> - How do we feel about the fact that, with this PR in, the name of this package now belies the extent of its functionality? It would now provides basic, heuristic-based code intel, but also precise code intel when LSIF is available.

I'll rethink this after GopherCon. Ideas:

- Rename it
- Move LSIF into lsp-client and depend on lsp-client
- Make a new package lsif, and have sourcegraph-{typescript,go} depend on all 3: lsif, basic-code-intel, and lsp-client

> This means that by default, language servers would essentially be prioritized over LSIF when available. This is the case on sourcegraph.com. Shouldn't it be the other way around?

Oh yeah, I completely overlooked this, thanks for pointing it out! I'll figure out how to set the priority to: LSIF, language server, then basic-code-intel.

package/src/lsp-conversion.ts
```diff
@@ -0,0 +1,81 @@
...
+    locationOrLocations: LSP.Location | LSP.Location[] | null
+): sourcegraph.Location[] | null {
+    if (!locationOrLocations) {
+        return null
+    }
+    const locations = Array.isArray(locationOrLocations)
+        ? locationOrLocations
+        : [locationOrLocations]
+    return locations.map(location => convertLocation(sourcegraph, location))
+}
```
felixfbecker: Why is this code copy pasted and not imported from lsp-client?

package/src/lsif.ts
```diff
@@ -0,0 +1,229 @@
...
+        sourcegraph.internal.sourcegraphURL
+    )
+    url.searchParams.set('repository', repositoryFromDoc(doc))
+    url.searchParams.set('commit', commitFromDoc(doc))
+
+    const response = await fetch(url.href, {
+        method: 'POST',
+        headers: new Headers({
+            'content-type': 'application/json',
+            'x-requested-with': 'Sourcegraph LSIF extension',
```
felixfbecker: This isn't accurate anymore, right?

package/src/lsif.ts
```diff
@@ -0,0 +1,229 @@
...
+    if (!response.ok) {
+        throw new Error(`LSIF /request returned ${response.statusText}`)
+    }
+    return await response.json()
+}
+
+export const mkIsLSIFAvailable = (lsifDocs: Map<string, Promise<boolean>>) => (
+    doc: sourcegraph.TextDocument,
+    pos: sourcegraph.Position
+): Promise<boolean> => {
```
felixfbecker: As a piece of feedback, due to the currying, abbreviated function name and no docblock, as a reviewer it took me a while to understand what this function does and how it is used (although I still don't understand why it is curried).

package/src/lsif.ts
```diff
@@ -0,0 +1,229 @@
...
+    url.searchParams.set('repository', repositoryFromDoc(doc))
+    url.searchParams.set('commit', commitFromDoc(doc))
+    url.searchParams.set('file', pathFromDoc(doc))
+
+    const hasLSIFPromise = (async () => {
+        try {
+            // Prevent leaking the name of a private repository to
+            // Sourcegraph.com by relying on the Sourcegraph extension host's
+            // private repository detection, which will throw an error when
+            // making a GraphQL request.
```
felixfbecker: This part of the code is not clear to me.
- How would we leak private repo names? Wouldn't we always contact the private instance?
- Why would the extension host throw an error on any GraphQL request _to the private instance_? How would it even intercept GraphQL requests?

package/src/lsif.ts
```diff
@@ -0,0 +1,229 @@
...
+    return convertLocations(
+        sourcegraph,
+        locations.map((reference: LSP.Location) => ({
+            ...reference,
+            uri: setPath(doc, reference.uri),
+        }))
+    )
+}
+
+export type Maybe<T> = { value: T } | undefined
```
felixfbecker: What is the benefit of this over a simple union type `T | undefined`?

template/src/extension.ts
```diff
@@ -1,66 +1,39 @@
...
     const languageID = 'all'
 
-    if (languageID === 'all') {
-        for (const languageSpec of languageSpecs) {
-            activateWithArgs(ctx, { ...languageSpec.handlerArgs, sourcegraph })
+    for (const languageSpec of languageID === 'all'
+        ? languageSpecs
+        : [languageSpecs.find(l => l.handlerArgs.languageID === languageID)!]) {
+        if (sourcegraph.configuration.get().get('codeIntel.lsif')) {
+            initLSIF()
```
felixfbecker: Doesn't `initLSIF()` return an object? What's the reason it is not used here?

package/src/lsif.ts
```diff
@@ -0,0 +1,229 @@
...
+    defaultR: R
+) => async (...args: A): Promise<R> => {
+    for (const f of fs) {
+        const r = await f(...args)
+        if (r !== undefined) {
+            return r.value
+        }
+    }
+    return defaultR
+}
```
felixfbecker: I find the above 4 helper functions are really hard to understand. Here's why: the function name could mean a lot of things, and there is no documentation, so one is left with reading the implementation to understand it - however, the code is very complex (generic and curried), and is hard to understand without seeing how it's being used, but the usages below don't make it a lot clearer either. I consider myself pretty good at TypeScript so if I have problems with this (>5min in it's still not clear to me) I assume anyone with less experience on our team would be left scratching their head. Please consider simplifying or at least documenting.

package/src/lsif.ts
```diff
@@ -0,0 +1,229 @@
...
+
+export function initLSIF() {
```
felixfbecker: Could it be that there is no TSLint setup in this repo? Because this function doesn't have a return type annotation, which would be rejected by our TSLint config (and also would have made it a lot easier to understand this code, as a reviewer)

.prettierignore
```diff
@@ -1,7 +1,8 @@
-node_modules/
-.git/
-coverage/
 .cache/
-lib/
+.nyc_output/
+coverage/
 dist/
-package.json
```
felixfbecker: package.json should not be formatted with prettier because yarn/npm override the formatting and format differently

package/src/lsp-conversion.ts
chrismwendt: https://github.com/sourcegraph/sourcegraph-basic-code-intel/blob/3abce29895c161bd6030dd2305158fa9acf2a6c6/package/src/lsp-conversion.ts#L1-L3

Feel free to commit a fix.

package/src/lsif.ts
chrismwendt: That's true, switching to `'Basic code intel'`

package/src/lsif.ts
chrismwendt: Added docs and refactored a bit.

package/src/lsif.ts
chrismwendt: Private repo names would have been leaked by a browser extension running on private GitHub.com repos.

There's already code in the extension host to detect if the page is a private repo. This logic triggers that code to get run at the expense of an extra GraphQL request. Do you have any ideas to improve this? Maybe the extension API could expose `isPrivateRepository`?

package/src/lsif.ts
chrismwendt: `{ value: T }` lets a consumer disambiguate between the two variants, and works for all `T` (including `undefined`).

template/src/extension.ts
chrismwendt: I think this is old code.

package/src/lsif.ts
chrismwendt: Added a bunch of docstrings and renamed vars

package/src/lsif.ts
chrismwendt: WIP on tslint in https://github.com/sourcegraph/sourcegraph-basic-code-intel/pull/125

PR #134, opened 2019-08-11
Replace `concurrently` with tsc-watch and a Procfile

chrismwendt:
- fs-events was causing high CPU https://github.com/sourcegraph/sourcegraph-basic-code-intel/issues/130
- I'm more comfortable with goreman and Procfiles than `concurrently`
- I couldn't get `concurrently` to show prefixes and color


chrismwendt:
https://www.npmjs.com/package/foreman says it shuts down the app when one process exits with a nonzero code, but it doesn't seem to actually do that. goreman shuts down properly.

I removed fsevents (by removing the only dependent package chokidar-infanticide-cli) and that didn't fix the CPU usage issue.

chrismwendt:
It's actually `parcel` which is using up lots of CPU. Parcel 2 might fix this https://medium.com/@devongovett/parcel-2-0-0-alpha-1-is-here-8b160c6e3f7e

felixfbecker:
But we use Parcel in sourcegraph/sourcegraph too and it doesn't cause high CPU there, so it can't be Parcel v1's fault.

Again, I think we _want_ to use `fsevents`. It is the native binding to the native macOS file watcher. _Not_ using `fsevents` causes high CPU usage, because it then needs to do polling.

My sense is still that dependencies are outdated. Did you try upgrading? What steps did you take?

chrismwendt:
I upgraded from Parcel 1.10.3 -> 1.12.3 and I think that added fsevents to yarn.lock and fixed the CPU usage problem.

To do that, I dropped all version constraints from package.json and completely regenerated yarn.lock, then added back a few relevant constraints.

chrismwendt:
I'd like to merge this in, any other comments?

chrismwendt:
@efritz Would you prefer tsc-watch and a Procfile (like the Sourcegraph app)?

template/package.json
```diff
@@ -109,24 +109,25 @@
...
+    "nyc": "",
+    "parcel": "1.x",
+    "prettier": "",
+    "source-map-support": "",
+    "sourcegraph": "23.0.1",
+    "ts-node": "",
+    "tsc-watch": "",
+    "tslint": "",
+    "tslint-language-service": "",
+    "typescript": ""
```
felixfbecker: ^ ranges are broken

felixfbecker:
I have no other comments than what I said above:

> Imo it's a little weird to require installation of Go and a Go tool to develop on / contribute to a TypeScript project.

If you do want to go with this, please add instructions to the README for contributors how to build the repo / install the needed tools.

PR #135, opened 2019-08-11
Add icons

chrismwendt:
image (20)

Resolves https://github.com/sourcegraph/sourcegraph/issues/5048

This inlines icons into the manifests so they appear in the extension registry.


generator/src/main.ts
```diff
@@ -101,6 +101,10 @@ function main(): void {
             'src/extension.ts'
         )
 
+        shell.exec(
+            `bash -c 'cat package.json | jq --argjson icon "$(echo -n "data:image/png;base64,$(base64 < ../icons/${languageID}.png)" | jq -R -s .)" ".icon = \\$icon" > package.json'`
+        )
```
felixfbecker: To me it seems like it would be a lot easier (esp. to understand) to just use `fs.readFile()` + `JSON.parse()` + `new Buffer().toString('base64')` + `JSON.stringify()`.

I personally have no idea how `jq` works, what the flags here mean etc.

nicksnyder:
It would be nice to include screenshots in the PR description for changes like this (I know that you shared them out of band too).

generator/src/main.ts
chrismwendt: Switched from `jq` to Node functions

PR #145, opened 2019-09-15
Interpret a non-OK response as no LSIF available

chrismwendt:
This is a quick fix because code intel appears to be broken right now.


lguychard:
Code change makes sense, although I'm curious as to why we haven't noticed this as an issue until now. Any ideas?

Should some response codes still be reported as errors?

chrismwendt:
The new LSIF SQLite backend might have caused this because it throws 500s when no LSIF data exists. @efritz is fixing the 500 in another PR, after which we can start logging some response codes as errors.

PR #157, opened 2019-10-05
add CUDA

beyang:
@chrismwendt I wasn't able to upload this to Sourcegraph.com, as I ran through the instructions (https://github.com/sourcegraph/sourcegraph-basic-code-intel#adding-a-new-sourcegraphsourcegraph-lang-extension-sourcegraphers-only), but couldn't get past the generator step. The error when I tried to run `bash generator.sh` was `Cannot find module './package/lib/handler'`. Is there a build step I'm missing that compiles to the `package/lib` folder?

Trying to make code intelligence work for a potential user @christinaforney has in mind on https://sourcegraph.com/github.com/gunrock/gunrock.


languages.ts
```diff
@@ -263,6 +263,39 @@ export const languageSpecs: LanguageSpec[] = [
         },
         stylized: 'C++',
     },
+    {
+        handlerArgs: {
+            languageID: 'cuda',
+            fileExts: ['cu', 'cuh'],
+            commentStyle: cStyle,
+            filterDefinitions: ({ filePath, fileContent, results }) => {
+                // Note: this is copy-pasted from C++
```
chrismwendt: I extracted this as a common function.

chrismwendt:
I added a build step to the generate script, should work now.

chrismwendt:
I also updated the docs to show arguments to the generator: `bash generator/generate.sh --languages cpp --publish`

PR #160, opened 2019-10-15
Decode the fragment

chrismwendt:
The problem here was that basic-code-intel was not decoding the URL fragment (the part after the `#`), which broke code intel on files with special characters.

Fixes https://github.com/sourcegraph/sourcegraph/issues/6042

### DIFF
--- a/package/src/api.ts
+++ b/package/src/api.ts
@@ -170,7 +170,7 @@ export class API {
             }
           }`
 
-        const { repo, rev, path } = parseUri(loc.uri.toString())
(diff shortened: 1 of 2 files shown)


package/src/api.ts
```diff
@@ -189,26 +189,12 @@ export class API {
 }
 
 export function parseUri(
-    uri: string
+    uri: URL
```
felixfbecker: 👍 for switching from `string` to `URL`

PR #171, opened 2019-11-18
Print log message when LSIF data is available

efritz:
This handles https://github.com/sourcegraph/sourcegraph/issues/6656.

### DIFF
--- a/package/src/lsif.ts
+++ b/package/src/lsif.ts
@@ -117,9 +117,7 @@ export const mkIsLSIFAvailable = () => {
             }
             const available: boolean = await response.json()
             if (available) {
-                console.log(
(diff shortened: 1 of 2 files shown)


package/src/lsif.ts
```diff
@@ -111,7 +115,13 @@ export const mkIsLSIFAvailable = () => {
             if (!response.ok) {
                 return false
             }
-            return await response.json()
+            const available: boolean = await response.json()
+            if (available) {
+                console.log(
+                    `LSIF data is available for repository=${repository} commit=${commit} file=${file}`
```
felixfbecker: This is still a bit ambiguous - it's available, but does that mean it's used?
-------

Would recommend logging an object `{ repository, commit, file }` because Chrome will highlight that and make it expandable

PR #178, opened 2019-12-04
LSIF GraphQL API

efritz:
This is an update to support the GraphQL API for LSIF data (once https://github.com/sourcegraph/sourcegraph/pull/7021 is merged).

The version check might be wonky: not sure a better way than to do it date-based. That will allow a transition period where both APIs are available (for at least one day). The date string there is currently wrong (obviously) and must be updated after the dependent PR is merged.

### DIFF
--- a/package/src/lsif.ts
+++ b/package/src/lsif.ts
@@ -127,7 +127,6 @@ export const mkIsLSIFAvailable = () => {
         })()
 
         lsifDocs.set(doc.uri, hasLSIFPromise)
-
         return hasLSIFPromise
     }
 }
@@ -146,11 +145,6 @@ export async function hover(
         return undefined
     }
 
-    console.debug('Received hover', {
-        source: 'HTTP API',
-        line: position.line,
-        character: position.character,
-    })
     return { value: convertHover(sourcegraph, hover) }
 }
 
@@ -171,12 +165,6 @@ export async function definition(
     if (locations.length === 0) {
         return undefined
     }
-
-    console.debug(`Received ${locations.length} definitions`, {
-        source: 'HTTP API',
-        line: position.line,
-        character: position.character,
-    })
 
     return {
         value: convertLocations(
@@ -203,12 +191,6 @@ export async function references(
     if (locations.length === 0) {
         return []
     }
-
-    console.debug(`Received ${locations.length} references`, {
-        source: 'HTTP API',
-        line: position.line,
-        character: position.character,
-    })
 
     return convertLocations(
         sourcegraph,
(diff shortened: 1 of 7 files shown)


package/src/lsif.ts
```diff
@@ -115,11 +120,7 @@ export const mkIsLSIFAvailable = () => {
             if (!response.ok) {
                 return false
             }
-            const available: boolean = await response.json()
-            if (available) {
-                console.log('Using LSIF data', { repository, commit, file })
```
chrismwendt: Why remove this indicator?

package/src/lsif.ts
```diff
@@ -286,6 +287,54 @@ export const noopMaybeProviders = {
...
+    if (productVersion === 'dev') {
+        return true
+    }
+
+    // Product versions have the form `49952_2019-12-04_c6f43e0`.
+    // We want to use the GraphQL resolvers if the instance is known to have them.
+    // Check to see if the product version occurs after the date the resolvers became
+    // available. We still support the LSIF HTTP API up to this date, so anything
+    // before it can still fall back safely.
+    return productVersion.split('_')[1].localeCompare(GRAPHQL_API_MINIMUM_DATE) > 0
```
chrismwendt: `productVersion` could also be something like `3.10.1` https://k8s.sgdev.org/api/console#%7B%22query%22%3A%22query%20%7B%20site%20%7B%20productVersion%20%7D%20%7D%22%7D

template/src/extension.ts
```diff
@@ -9,10 +9,14 @@ export function activate(ctx: sourcegraph.ExtensionContext = DUMMY_CTX): void {
     // This is set to an individual language ID by the generator script.
     const languageID = 'all'
 
+    // LSIF is not language-specific, and we only want to initialize it once.
+    // Otherwise we will make a flurry of calls to the frontend to check if
+    // LSIF is enabled.
+    const lsif = initLSIF()
```
chrismwendt: :+1: and FYI this template extension which bundles all languages only runs during development.

package/src/lsif.ts
chrismwendt: To be safe, let's run a regex on this to ensure it's the expected format before asserting that the instance supports GraphQL.

efritz: Yeah I'd like to discuss the possibilities of this a bit more. Quick in-person chat while we're both in SF?

package/src/lsif.ts
efritz: It was no longer symmetric with GraphQL, which didn't have a way to determine if there was a dump without actually making the query first.

After changes to RFC 74, this will be reworked again.

package/src/lsif.ts
efritz: Updated to deal with semantic versions as well - this might be a good enough indicator now, but we should still discuss it.

package/src/lsif.ts
efritz: I added logging for each request instead.

PR #192, opened 2020-01-22
Update typescript versions.

efritz:
This matches the current typescript version in sg/sg.

### DIFF
--- a/package/src/memoizeAsync.ts
+++ b/package/src/memoizeAsync.ts
@@ -5,7 +5,7 @@
  * @param resolver If resolver provided, it determines the cache key for storing the result based on
  * the first argument provided to the memoized function.
  */
-export function memoizeAsync<P extends object, T>(
+export function memoizeAsync<P extends {}, T>(
     func: (params: P) => Promise<T>,
     resolver?: (params: P) => string
 ): (params: P, force?: boolean) => Promise<T> {
--- a/package/src/memoizeAsync.ts
+++ b/package/src/memoizeAsync.ts
@@ -5,7 +5,7 @@
  * @param resolver If resolver provided, it determines the cache key for storing the result based on
  * the first argument provided to the memoized function.
  */
-export function memoizeAsync<P, T>(
+export function memoizeAsync<P extends object, T>(
     func: (params: P) => Promise<T>,
     resolver?: (params: P) => string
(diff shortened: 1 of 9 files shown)


package/src/memoizeAsync.ts
```diff
@@ -5,7 +5,7 @@
  * @param resolver If resolver provided, it determines the cache key for storing the result based on
  * the first argument provided to the memoized function.
  */
-export function memoizeAsync<P, T>(
+export function memoizeAsync<P extends object, T>(
```
chrismwendt: Not sure if this was intentional, but it looks harmless.

chrismwendt (approved):

package/src/memoizeAsync.ts
efritz: Yes, toString is not defined otherwise. Causes error in new ts.

felixfbecker: I think this is not valid, because `memoizeAsync` also works fine with scalars. Using `{}`  would match scalars too I think.

efritz: How does this work with scalars?

https://github.com/sourcegraph/sourcegraph-basic-code-intel/blob/361b25d89c6cae917e86825b0817d0a23a34006a/package/src/memoizeAsync.ts#L14

felixfbecker: If params is a number, string, or boolean, `params.toString()` will return the decimal string, the string itself, or "true"/"false" respectively.

efritz:
```typescript
src/memoizeAsync.ts:14:58 - error TS2339: Property 'toString' does not exist on type 'P'.

14         const key = resolver ? resolver(params) : params.toString()
```

felixfbecker: I don't know what code that error came from, but it works in the playground:
https://www.typescriptlang.org/play/#code/GYVwdgxgLglg9mABMAPABUQUwB5U2AEwGdEBvAXwD4AKABwC5E0BKMgKEU8VoDoo4AylABOMMAHNqzNuSA

efritz: OH NO. I removed extends completely. I got it now.

felixfbecker (approved):

PR #193, opened 2020-01-22
Emit telemetry events

efritz:
This PR emits code intel telemetry events after receiving a result from a provider and directly before yielding it to the UI. This emits the time until the _first_ event of that type. If there are multiple pages of LSIF results, then only the time of the first one to load will emit an event.


template/src/extension.ts
```diff
@@ -19,6 +19,25 @@ export function activate(ctx: sourcegraph.ExtensionContext = DUMMY_CTX): void {
     // LSIF is enabled.
     const lsif = initLSIF()
 
+    // Emit a telemetry event with a durationMs attribute. The value of
+    // `actionStartedAt` should be the timestamp taken at the beginning
+    // of the action. All telemetry events are prefixed with 'codeintel'.
+    const log = async (
+        actionStartedAt: number,
```
chrismwendt: Another idea: `log` could take a callback and measure timing itself.

shared/telemetry.ts
```diff
@@ -0,0 +1,53 @@
+import * as sourcegraph from 'sourcegraph'
+
+// Polyfill
+const g: any = global
+g.performance = g.performance || { now: () => 0 }
```
chrismwendt: Out of curiosity, why use define `const g` instead of modifying global directly?

chrismwendt: Could `Date.now()`be used as a fallback? They both return ms.

chrismwendt (approved):

PR #197, opened 2020-01-24
Enable search-based intelligence for Objective C.

### LINKED ISSUE #6655: Basic intellisense for objective-C
#### Question description

Is objctive C supported by sourcegraph, it seems that j2d does not work in `.m` files.

#### Additional context

https://sourcegraph.com/github.com/changsanjiang/SJVideoPlayer@daef27455ee9ef71945bd323227beb51dc329086/-/blob/Example/1%20ObjC%20Example/SJVideoPlayerExample-ObjC/Pods/Masonry/Masonry/MASCompositeConstraint.m

creachadair:
Since Objective C shares header file extensions with C and C++, this change treats Objective C as a dialect of C++ for the purpose of basic code intelligence. This is a mild (and arguably reasonable) compromise to the convention that each language has its own extension: We already treat C and C++ as one language, and since Objective C subsumes native C syntax as well, it makes sense to allow them to share filtering logic.

The substance of this change is to add the .m file extension to the C++ extension, and to extend the import filter to allow #import and @import directives as well as #include. For (correct) native C and C++ code, these extensions will have no material effect.

Fixes https://github.com/sourcegraph/sourcegraph/issues/6655.

### DIFF
--- a/languages.ts
+++ b/languages.ts
@@ -40,6 +40,17 @@ const lispStyle: CommentStyle = {
     },
 }
 
+/**
+ * Filter a list of candidate definitions to select those likely to be valid
+ * cross-references for a definition in this file. Accept candidates located in
+ * files that are a suffix match (ignoring file extension) for some import of
+ * the current file.
+ *
+ * For imports we examine user `#include` and `#import` paths, as well as
+ * Objective C module `@import` package names. If no candidates match, fall
+ * back to the raw (unfiltered) results so that the user doesn't get an empty
+ * response unless there really is nothing.
+ */
 const cppFilterDefinitions: HandlerArgs['filterDefinitions'] = ({
     filePath,
     fileContent,
@@ -48,17 +59,28 @@ const cppFilterDefinitions: HandlerArgs['filterDefinitions'] = ({
     const imports = fileContent
         .split(/\r?\n/)
         .map(line => {
-            const match = /^#include "(.*)"$/.exec(line)
-            return match ? match[1] : undefined
+            // Rewrite `@import x.y.z;` as x/y/z to simulate path matching.
+            // In plain C and C++ files, expect this to be empty.
+            const matchImport = /^@import (\w+);$/.exec(line)
+            if (matchImport) {
+                return matchImport[1].replace(/\./g, '/')
+            }
+
+            // Capture paths from #include and #import directives.
+            // N.B. Only user paths ("") are captured, not system (<>) paths.
+            return /^#(include|import) "(.*)"$/.exec(line)?.[2]
         })
         .filter((x): x is string => Boolean(x))
 
+    // Select results whose file path shares a suffix with some import.
+    // N.B. Paths are compared without file extensions.
     const filteredResults = results.filter(result => {
-        return imports.some(i =>
-            path
-                .join(path.parse(result.file).dir, path.parse(result.file).name)
-                .endsWith(path.join(path.parse(i).dir, path.parse(i).name))
-        )
+        const resultParsed = path.parse(result.file)
+        const candidate = path.join(resultParsed.dir, resultParsed.name)
+        return imports.some(i => {
+            const iParsed = path.parse(i)
+            return candidate.endsWith(path.join(iParsed.dir, iParsed.name))
+        })
     })
 
     return filteredResults.length === 0 ? results : filteredResults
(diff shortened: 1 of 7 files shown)


chrismwendt (approved):
It needs more thought, but there might be another way to solve this that's more in line with having 1 extension for each language, e.g. by running both the CPP and Objective C extensions and checking the file content to determine which language it is.

languages.ts
```diff
@@ -258,6 +281,7 @@ export const languageSpecs: LanguageSpec[] = [
                 'h',
                 'hpp',
                 /* Arduino */ 'ino',
+                /* Objective C */ 'm',
```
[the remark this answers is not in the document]
creachadair: You can, but then .h files don't go the right place. Sniffing content as you suggested above might be a way around that, but there's a rathole that I'm not convinced is worthwhile. 

There's a reasonable case that Objective C is not really a distinct language: It subsumes most of C (or C++ for Objective C++) and in practice the same compiler toolchains build both (Clang, for example uses the same code paths in the front end).

With that in mind, do we gain anything besides aesthetic symmetry from a "1 extension per langauge" rule?

languages.ts
```diff
@@ -48,17 +59,29 @@ const cppFilterDefinitions: HandlerArgs['filterDefinitions'] = ({
...
+            // Rewrite `@import x.y.z;` as x/y/z to simulate path matching.
+            // In plain C and C++ files, expect this to be empty.
+            const matchImport = /^@import (\w+);$/.exec(line)
+            if (matchImport) {
+                return matchImport[1].replace(/\./g, '/')
+            }
+
+            // Capture paths from #include and #import directives.
+            // N.B. Only user paths ("") are captured, not system (<>) paths.
+            const matchInclude = /^#(include|import) "(.*)"$/.exec(line)
```
felixfbecker:
```suggestion
            return /^#(include|import) "(.*)"$/.exec(line)?.[2]
```

languages.ts
```diff
@@ -48,17 +59,29 @@ const cppFilterDefinitions: HandlerArgs['filterDefinitions'] = ({
...
+            return matchInclude ? matchInclude[2] : undefined
```
felixfbecker:
```suggestion
```

languages.ts
```diff
@@ -48,17 +59,29 @@ const cppFilterDefinitions: HandlerArgs['filterDefinitions'] = ({
...
 
+    // Select results whose file path shares a suffix with some import.
+    // N.B. Paths are compared without file extensions.
     const filteredResults = results.filter(result => {
-        return imports.some(i =>
-            path
-                .join(path.parse(result.file).dir, path.parse(result.file).name)
-                .endsWith(path.join(path.parse(i).dir, path.parse(i).name))
-        )
+        const rp = path.parse(result.file)
```
felixfbecker: Would you mind giving this a written-out variable name? Maybe it's just me but I'm not sure what `rp`/`ip` are supposed to mean 🙇

languages.ts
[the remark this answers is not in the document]
chrismwendt: It could work both ways, so let's go with how you've implemented it now. We can always revisit this if an issue arises.

languages.ts
creachadair: Done. Their scope is small enough and their usage obvious enough I don't think it really helps legibility any, but I am not picky about it.

languages.ts
creachadair: I really liked this suggestion, but `prettier` does not. Specifically, it fails our format check in CI. Do you have a suggestion how I can tweak the formatter or is it just that it doesn't recognize this syntax?

felixfbecker: Looks like this repo's dependencies have been a bit neglected: https://github.com/sourcegraph/sourcegraph-basic-code-intel/pulls?q=is%3Apr+is%3Aopen+author%3Aapp%2Frenovate+sort%3Aupdated-desc

There's an open PR to update Prettier: https://github.com/sourcegraph/sourcegraph-basic-code-intel/pull/175
It just needs a run of `yarn prettier` to reformat files with the new version.

If you want, you can get that PR merged or apply the update on this branch.

languages.ts
felixfbecker: Thanks, I understand now what you meant!

creachadair:
I merged #175 into this PR (with distinct commits) and cleaned up the history. CI is happy again.

felixfbecker (approved):
✨

creachadair:
Any objections to me merging this with the affiliated (but not directly related) formatting changes?

chrismwendt:
> Any objections to me merging this with the affiliated (but not directly related) formatting changes?

Not at all, that's fine, especially because the formatting is isolated to one commit.

PR #200, opened 2020-01-28
Add basic code intelligence for Pascal, Verilog, and VHDL

### LINKED ISSUE #4915: Add Verilog, Systemverilog, and VHDL language extension
#### Feature request description

From a user:
> I've looked at sourcegraph before but it seems to have had some of churn on the language support issue.  Previously the basic-code-intell worked out of the box for these languages (Verilog, Systemverilog and VHDL).  Now, this useful feature has been dropped.  I dug a little and see that it might be possible for me to add these languages but I'm afraid the update to the basic-code-intell stuff might be just out of my reach.  The doc on that repo is a bit confusing.  If I read it right , basic code intell uses universal-ctags under the hood (which I've used before) but I don't quite understand how to configure it on a per-language basis.  

Seems like .v, .vh, .svh, and .vhdl are the file extensions covered here...

creachadair:
For the results of this change to be visible in the UI, you must also sync Sourcegraph past https://github.com/sourcegraph/sourcegraph/pull/8079.

Fixes https://github.com/sourcegraph/sourcegraph/issues/4915.

### DIFF
--- a/languages.ts
+++ b/languages.ts
@@ -552,4 +552,38 @@ export const languageSpecs: LanguageSpec[] = [
         },
         stylized: 'R',
     },
+    {
+        handlerArgs: {
+            languageID: 'pascal',
+            fileExts: ['p', 'pas', 'pp'],
+            commentStyle: {
+                // Traditional: (* this is a comment *)
+                // Customary:   { this is also a comment }
+                block: {
+                    startRegex: /(\{|\(\*)\s?/,
+                    endRegex: /(\}|\*\))/,
+                },
+
+                // TODO: Some Pascal implementations support //-comments too.
+                // Is that common enough to support here?
+            },
+        },
+        stylized: 'Pascal',
+    },
+    {
+        handlerArgs: {
+            languageID: 'verilog',
+            fileExts: ['sv', 'svh', 'svi', 'v'],
+            commentStyle: cStyle,
+        },
+        stylized: 'Verilog',
+    },
+    {
+        handlerArgs: {
+            languageID: 'vhdl',
+            fileExts: ['vhd', 'vhdl'],
+            commentStyle: { lineRegex: /--+\s?/ },
+        },
+        stylized: 'VHDL',
+    },
 ]
--- a/samples/pascal.pas
+++ b/samples/pascal.pas
@@ -1,0 +1,676 @@
+unit bzip2;
+{****************************************************************************
+
+                             BZIP2 decompression unit
+
+                        Copyright (C) 2002 by Daniel Mantione
+
+This unit provides a decompression stream to decode .bz2 files. It is
+inpired by Julian R. Seward's libbzip2 library and therefore you should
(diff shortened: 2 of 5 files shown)


languages.ts
```diff
@@ -552,4 +552,38 @@ export const languageSpecs: LanguageSpec[] = [
         },
         stylized: 'R',
     },
+    {
+        handlerArgs: {
+            languageID: 'pascal',
```
chrismwendt: When running locally, Pascal code intel doesn't work. I suspect it's because there's no Pascal entry in https://github.com/sourcegraph/sourcegraph/blob/3a1dd8ccd49175ef2ed47dabe85d2e0543ce6c47/shared/src/languages.ts#L40

languages.ts
```diff
@@ -552,4 +552,38 @@ export const languageSpecs: LanguageSpec[] = [
...
+        handlerArgs: {
+            languageID: 'verilog',
+            fileExts: ['sv', 'svh', 'svi', 'v'],
+            commentStyle: cStyle,
+        },
+        stylized: 'Verilog',
+    },
+    {
+        handlerArgs: {
+            languageID: 'vhdl',
```
chrismwendt: Also needs an entry in https://github.com/sourcegraph/sourcegraph/blob/3a1dd8ccd49175ef2ed47dabe85d2e0543ce6c47/shared/src/languages.ts#L40

chrismwendt (approved):
While trying this out I found some samples:

- Verilog https://sourcegraph.com/github.com/ZipCPU/wbuart32/-/blob/bench/verilog/helloworld.v
- VHDL https://sourcegraph.com/github.com/xesscorp/VHDL_Lib@master/-/blob/HostIoComm.vhd

It's ok to merge this in before adding language entries to https://github.com/sourcegraph/sourcegraph/blob/3a1dd8ccd49175ef2ed47dabe85d2e0543ce6c47/shared/src/languages.ts#L40

languages.ts
creachadair: Yes, in my local testing I also have https://github.com/sourcegraph/sourcegraph/pull/8079 patched in. That _should_ give you working results, but please let me know if it does not.

languages.ts
creachadair: Yes, thanks, that's part of https://github.com/sourcegraph/sourcegraph/pull/8079.

creachadair:
> * Verilog https://sourcegraph.com/github.com/ZipCPU/wbuart32/-/blob/bench/verilog/helloworld.v

Thanks, your Verilog example was nicer than the one I was using.

chrismwendt (approved):

mpettigr:
Here's an ethernet mac written in verilog if needed. 
https://github.com/freecores/ethmac/tree/master/rtl/verilog 

Bunch of different VHDL library elements
https://github.com/xesscorp/VHDL_Lib

PR #205, opened 2020-01-28
Add icons for Pascal, Verilog, and VHDL.

creachadair:
This is a follow-up to #200. Publishing extensions requires an icon for each specified language.

As far as I can tell, there are no "official" logos for Pascal, Verilog, or VHDL. There are commercial products that have logos, but those go with the product rather than the language. So here I have opted for the low-budget expedient of creating flat square name icons.


efritz (approved):

efritz:
The verilog baseline looks a bit higher than the other two.

creachadair:
> The verilog baseline looks a bit higher than the other two.

Probably. I had to make the typeface smaller for that one, and probably didn't get the alignment quite identical.

PR #224, opened 2020-02-03
Clojure: add slash `/` to recognized characters

chrismwendt:
Requested at https://github.com/sourcegraph/sourcegraph/issues/3614#issuecomment-581636768

**Edit** Hang on, I might have misinterpreted the request as relating to symbol names instead of definition keywords.


chrismwendt:
This is wrong. `/` separates a namespace from a symbol e.g. https://sourcegraph.com/github.com/plumatic/schema/-/blob/src/clj/schema/experimental/generators.clj#L64

PR #229, opened 2020-02-13
Paginate LSIF references

efritz:
This modifies the LSIF references provider to read subsequent pages if an endCursor is given in the first request. This will enable cross-repo LSIF results to be shown in the UI.


creachadair (approved):

chrismwendt (approved):
I'd set a limit on the page count, and it's ok to be redundant if there's a server-side limit.

efritz:
@chrismwendt Updated could you take another quick look?

chrismwendt (approved):

PR #230, opened 2020-02-13
Refactor language definitions

efritz:
- Reorganize files
- Add tests for filterDefinitions
- Rewrite filter definitions to be self-documenting
- Prune the Result interface of unnecessary items (moved search result and additional fields that filter definitions doesn't really care about into search package)

A few changes to filter definitions:
- java now supports static wildcard imports
- typescript now supports importing an index.ts/js file


shared/language-specs/common.ts
```diff
@@ -0,0 +1,35 @@
+import { CommentStyle } from './spec'
```
creachadair: [minor nit] "common.ts" seems a little generic. What do you think about "comments.ts" to focus it on comment styles rather than suggesting a grab bag?

shared/language-specs/languages.ts
```diff
@@ -624,10 +277,17 @@ export const languageSpecs: LanguageSpec[] = [
     },
 ]
 
+/**
+ * Returns the language spec with teh given language identifier. If no language
```
creachadair:
```suggestion
 * Returns the language spec with the given language identifier. If no language
```

shared/language-specs/languages.ts
```diff
@@ -624,10 +277,17 @@ export const languageSpecs: LanguageSpec[] = [
...
+ * matches is configured with the given identifier an error is thrown.
+ *
+ * @param languageID The language ID.
```
creachadair: [not blocking] I'm not fond of explicit `@param` docs that add nothing beyond what the parameter declaration itself says. They're actually _worse_ than nothing, because they add mass to the file, can get out of sync, and force the reader to explicitly pay attention to a triviality.

I'd vote for dropping this entirely, or including it only if we have something to add about the format of the string (e.g., a pointer to the canonical language ID docs or something).

shared/language-specs/python.ts
```diff
@@ -0,0 +1,84 @@
...
+    return importPath.replace(/\./g, '/')
+}
+
+/**
+ * Converts a Python import path into a file path relative to the
+ * given source path. If the import path is not relative, method
+ * function returns undefined.
+ *
+ * @param sourcePath The source file.
+ * @param importPath The relative or absolute Python import path.
```
creachadair:
```suggestion
 * @param importPath The relative or absolute Python import path (`.a.b`, `a.b.c`).
```

shared/language-specs/python.ts
```diff
@@ -0,0 +1,84 @@
...
+function absoluteImportPath(importPath: string): string {
+    return importPath.replace(/\./g, '/')
+}
+
+/**
+ * Converts a Python import path into a file path relative to the
+ * given source path. If the import path is not relative, method
+ * function returns undefined.
+ *
+ * @param sourcePath The source file.
```
creachadair:
```suggestion
 * @param sourcePath The source file path relative to the repository root.
```

(I think this is correct)

shared/language-specs/spec.ts
```diff
@@ -0,0 +1,155 @@
+/**
+ * The specification used to provide search-based code intelligence for a
+ * particular language. This includes things like file extensions, comment
+ * patterns and delimiters, and logic for filtering out obviously wrong
+ * search results for definitions.
+ */
+export interface LanguageSpec {
+    /**
```
creachadair: I find the vertical displacement distracting here. Can we make one-liners of the ones that have only one line? E.g.,

```suggestion
/** Used to label markdown code blocks */
```

I'm pretty sure the doc comment parser is fine with this form.

shared/language-specs/spec.ts
```diff
@@ -0,0 +1,155 @@
...
+     * The path to the result file relative to the repository root.
+     */
+    file: string
+}
+
+/**
+ * A zero-value filter context used for testing.
+ */
+export const nilFilterContext: FilterContext = {
```
creachadair: Can `nilFilterContext` and `nilResult` live in a test-only file instead?  (I'm assuming we don't ship `.test.ts` files in prod builds)

shared/language-specs/util.ts
```diff
@@ -0,0 +1,81 @@
+import { isDefined } from '../util/util'
+import { Result } from './spec'
+
+/**
+ * Extract content from each line of the source file. The first match on
+ * each line is returned. Each supplied pattern is expected to have exactly
```
creachadair: [optional] Maybe this can be a separate change, but we could make this more robust by taking the whole match if there isn't a capture group defined (e.g., "The first capture group of each pattern is returned on a successful match; or if there are no capture groups the whole match is returned." or words to that effect)

```ts
return match.length > 1 ? match[1] : match[0]
```

or whatever.

shared/search/conversion.ts
```diff
@@ -1,25 +1,42 @@
...
-        ),
-        range: new sourcegraph.Range(
-            result.start.line,
-            result.start.character,
-            result.end.line,
-            result.end.character
-        ),
-    }
+export interface Result {
+    /**
```
creachadair: As above, can we compact these one-liners a bit?

```suggestion
/** The name of the repository containing the result. */
```

et seq.

creachadair (approved):
I like this change, and I really appreciate the tests. I have a few minor comments, but nothing crucial.

chrismwendt (approved):

shared/language-specs/common.ts
efritz: Renamed.

shared/language-specs/util.ts
efritz: I don't think we need that fallback logic right now. I don't see an easy import syntax in the future that would match the entire string and not require a capturing group. Currently I think the docs explain what is expected of its input.

shared/language-specs/spec.ts
efritz: Updating it.

PR #232, opened 2020-02-14
Deduplicate regex patterns in language specs.

efritz:


creachadair (approved):

shared/language-specs/comments.ts
```diff
@@ -1,35 +1,46 @@
...
+/** Matches three slashes followed by one optional space. */
+export const tripleSlashPattern = /\/\/\/\s?/
+
+/** Matches a hash followed by one optional space. */
+export const hashPattern = /#\s?/
+
+/** Matches two or more dashes followed by one optional space. */
+export const dashPattern = /---*\s?/
+
+// TODO - does not need to be optional?
```
creachadair: I wondered about this too. I would have expected this to be something like `^\s*(\*\s?)?`, so that it will match and consume star noise if it's present but also accept lines (given to be inside a block comment) that lack star noise.

I wasn't able to convince myself I wouldn't break something by changing it, but I think we only ever use it for matching rather than trimming, so it is _probably_ fine.

efritz: Will look into this a bit more.

PR #233, opened 2020-02-14
Support multiple comment styles per language.

efritz:
Some languages like rust and haskell have particular comment styles that are legal only above or below a definition, but both are defined in the language.

`-- | is a comment above in haskell`

`-- ^ is a comment below in haskell`

`//! documents the containing scope in rust`

`/// documents the next declaration in rust`

### DIFF
--- a/shared/language-specs/languages.ts
+++ b/shared/language-specs/languages.ts
@@ -1,10 +1,9 @@
 import {
-    cStyleBlockComment,
     cStyleComment,
     dashPattern,
-    hashPattern,
-    leadingAsteriskPattern,
+    javaStyleComment,
     leadingAtSymbolPattern,
+    leadingHashPattern,
     lispStyleComment,
     pythonStyleComment,
     shellStyleComment,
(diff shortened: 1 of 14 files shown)


shared/search/docstrings.ts
```diff
@@ -13,64 +13,67 @@ import {
...
+                      }
+
+            const blockCommentDocstring = findDocstringInBlockComment({
+                block: modifiedBlock,
+                lines: mungeLines(allLines, docPlacement, definitionLine),
+                docstringIgnore,
+            })
+            if (blockCommentDocstring) {
+                return unmungeLines(blockCommentDocstring, docPlacement).join(
+                    '\n'
```
creachadair: Did the formatter really do this here and elsewhere?

shared/language-specs/languages.ts
```diff
@@ -23,128 +22,149 @@ const clojureSpec: LanguageSpec = {
...
+            // -- | doc comment
+            // {- block comment -}
+            lineRegex: /--\s?\|?\s?/,
+            block: { startRegex: /{-/, endRegex: /-}/ },
+            docstringIgnore: haskellPragma,
+        },
+        {
+            // -- ^ doc comment
+            lineRegex: /--\s?\^?\s?/,
+            docPlacement: 'above the definition',
```
creachadair: I think this is backward, isn't it? 

```haskell
f :: A -> B
-- ^Do a thing (after)
```

Or does "above" mean relative to the comment?

creachadair (approved):
Really just one question, and one grumble about formatting that probably isn't your fault.

shared/language-specs/languages.ts
efritz: This is why we have code reviews.

shared/search/docstrings.ts
efritz: Yes it hits line lengths and that’s where it decided to break. Not much control over it, sometimes it’s very stupid.

PR #234, opened 2020-02-14
Add test to ensure icon files exist for every defined language.

efritz:


shared/language-specs/icons.test.ts
```diff
@@ -0,0 +1,34 @@
...
+                        exists: await fs.exists(filePath),
+                    }
+                })
+            )
+        )
+            .filter(({ exists }) => !exists)
+            .map(({ languageID }) => languageID)
+
+        if (languageIDs.length > 0) {
+            assert.fail(`Missing icons for ${languageIDs.join(',')}.`)
```
creachadair: Can we make it so that when this fails, it also logs instructions for how to fix it? E.g.,

`Each LANG must have a corresponding icons/LANG.png file`

so you don't have to read the source of the test to figure out what to fix?

creachadair (approved):
Thanks for adding this. :)

ryan-blunden (approved):

shared/language-specs/icons.test.ts
efritz: Done.

PR #239, opened 2020-02-17
Refactor query generation. Ensure pattern type is regexp.

efritz:


creachadair (approved):

rvantonder:
I think striving for a test for this functionality would be extremely useful. I'm not asking for it in this PR, 
but I'm kind of surprised we don't have something that validates expected behavior.

The web e2e tests are kind of a pain, so something I've been wondering about is whether we could set up an e2e test `src-cli`. Thoughts?

rvantonder:
Also can we update the description to say something like "Ensure pattern type is regexp for find-reference" queries or the like? It's unclear this change is related to the find-references issue, or other broader code intel functionality.

efritz:
> Also can we update the description (and commit) to say something like "Ensure pattern type is regexp for find-reference queries" or the like? It's unclear this change is related to the find-references issue, or other broader code intel functionality.

Fun thing, I think that issue was a search timeout on a monorepo and not actually this. I'm just taking the effort to make the code less dumb here (and covering the case when it won't be treated as a regex in some environments).

efritz:
> something I've been wondering about is whether we could set up an e2e test with src-cli.

What would this look like? I'm looking forward to the RFC 😛

rvantonder:
> What would this look like?

Basically what we do for e2e tests now, but without spinning up a web browser/chromium.

> Fun thing, I think that issue was a search timeout on a monorepo and not actually this. 

OK, probably worth noting in the parent issue.

PR #240, opened 2020-02-18
Fallback to indexed search on timeouts

efritz:
This PR aims to help some customers that have reported always having an empty locations panel when finding references without a precise provider (LSIF or LSP).

Find references performs two queries with the following repo scoping terms:

-  ```repo:^{repo}$@commit```
- ```-repo:^{repo}$```

The first finds the references in the current git tree. The second finds all references outside of the repo. The second one will use the Zoekt index naturally, but the first query only uses the Zoekt index if that's the currently indexed commit (almost always HEAD).

For customers with monorepos, the current search queries are very problematic. The first query will never find any results with non-indexed search as it takes too long to load the current commit into memory to perform a search. The second query will never find any results as this set of repos is empty (everything not the monorepo is nothing, or at the very least unlikely to contain the correct references).

This PR breaks the first type of query up into several fallback requests. First, launch a query with ```repo:^{repo}$@commit```. If there are no results within _N_ms, we start another query with ```repo:^{repo}$ index:only```. This will force an index search on whatever commit is available. The two queries then race and we return the first to resolve.

Additionally, we add the following two configuration values:

- `basicCodeIntel.indexOnly`, which turns off any unindexed search query to provide code intel in the UI
- `basicCodeIntel.unindexedSearchTimeout`, which lowers/raises the default 5s timeout after which an index-only search will be performed


shared/search/providers.ts
```diff
@@ -249,23 +238,154 @@ function getFileContent(
...
     }
 
-    return (await searchViaApi(query, shouldRequestFileLocal())).flatMap(
+    return (await searchViaApi(query, getConfig('fileLocal', false))).flatMap(
         searchResultToResults
     )
 }
 
+/**
+ * Filter search results before passing them to the language spec filter function.
```
creachadair: This comment is partly pre-existing, but I think we can improve it by talking about what the function does rather than how it's used. What I _think_ this means is something like:

```typescript
// Report whether the given symbol search result is visible to the document
// with the given text and file path. A symbol is visible if it is heuristically
// public or is defined in the file named by path.
```

I am not completely sure of this wording (e.g., is it "defined" or "located"? Is it "visible TO" or "visible FROM"?) but I'd like to focus the comment more on what it means for `isSymbolVisible(doc, path, sym)` to be true.

shared/search/providers.ts
```diff
@@ -249,23 +238,154 @@ function getFileContent(
...
+
+/**
+ * Perform a search query for references. Returns results converted to locations.
+ * Results are not sorted in any meaningful way as these results are meant to be
+ * merged with other search query results.
+ *
+ * @param args Parameter bag.
+ */
+async function searchReferences({
+    query,
```
creachadair: This seems like a lot of ceremony to pass in a query string. Given this isn't public, can we just make this `async function searchReferences(query: string): Promise<sourcegraph.Location[]> {` ?

(I get extensibility, but we own all the call sites)

creachadair (commented):
This approach makes sense (and in fact I had expected this is already what we were doing). I have a few comments, but overall I like it.

shared/search/tokens.ts
```diff
@@ -55,12 +55,17 @@ export function findSearchToken({
...
     const searchToken = line.substring(start, end)
 
-    if (
-        !lineRegexes.some(
-            // comment on this line occurring before the token
-            lineRegex => (line.match(lineRegex)?.index || 0) < start
-        )
-    ) {
+    // Determine if the token occurs after a comment on the same line
+    const insideComment = lineRegexes.some(lineRegex => {
```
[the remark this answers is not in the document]
efritz: This was an unrelated bug. I accidentally wrote it so that if there was no comment on the line it would think it was still in a comment.

shared/search/providers.ts
efritz: I think it can be even simpler.

```
Filter out search results that are a non-public symbol that does not belong to the current text document.
```

shared/search/providers.ts
efritz: We can't. It was written this way so that it could be passed to `searchWithFallback`, which has a generic input parameter extending `{query: string}`.

shared/search/promises.test.ts
```diff
@@ -0,0 +1,14 @@
+import * as assert from 'assert'
+import { getFirst } from './promises'
+
+describe('getFirst', () => {
+    it('returns first non-empty array', async () => {
+        const results = await getFirst(
+            Promise.resolve([]),
+            Promise.resolve([1, 2, 3]),
+            Promise.resolve([4, 5, 6])
```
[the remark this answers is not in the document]
efritz: Updated docs of tested function.

shared/search/providers.ts
```diff
@@ -152,23 +130,34 @@ export function createProviders({
...
+            query,
         })
 
-        // Perform all search queries concurrently, then merge and sort
-        // the results based on the proximity to the current file.
-        const searchResults = await Promise.all(queries.map(search))
-        const mergedResults = flatten(searchResults).map(resultToLocation)
-        return sortByProximity(mergedResults, new URL(doc.uri))
+        // Perform an indexed search over all _other_ repositories. This
+        // query is ineffective on DotCom as we do not keep repositories
```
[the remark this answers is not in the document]
efritz: No, this query should use the zoekt index for all repos that don't match a particular name. For orgs with monorepos, this should basically return an empty set of repos and perform no search. For dotcom, the gitserver is mostly ephemeral so the `-repo` term basically selects **all** of github. For other instances, this should be a health selection of repos that actually are indexed.

shared/search/providers.ts
creachadair: I agree that's simpler, but it only addresses part of what I was hoping to convey:

1. This function does not filter anything, it is a predicate. I was hoping to get the comment to describe what the predicate _means_, rather than what it's used for.
2. The name seems not to match the semantics: It's not really about visibility so much as location. Maybe something like `isExternalPrivateSymbol` or something.

shared/search/promises.ts
```diff
@@ -0,0 +1,54 @@
+/**
+ * Evaluate each promise sequentially. Return the result of the promise
+ * to yield a non-empty set of results. The result of subsequent promises
+ * will be ignored.
+ *
+ * @param ps A list of in-flight promises.
+ */
+export async function getFirst<T>(...ps: Promise<T[]>[]): Promise<T[]> {
```
creachadair: The comment update helps, thanks. 

I still think this shouldn't be exported as a library function. We only need it for one very specific case. I think it should be unexported and made part of that library, until we discover we need it for something else. It could be simplified too, since we only ever care about two:

```typescript
async function getFirst<T>(p: Promise<T[]>, q: Promise<T[]>): Promise<T[]> {
   const values = await p
   if (values.length !== 0) {
      return values
   }
   return await q
}
```

or words to that effect.

This isn't a blocker, though.

creachadair (approved):
My remaining comments are non-blocking.

shared/search/promises.ts
efritz: FYI, you cant test it if you can't import it.

And I don't see how a variadic function is much more complex than something that takes exactly two, especially if the body of the function is going to treat them as a sequential list. I wrote this function specifically this way because I can see us possibly needing to wait on a third value if we need to change the fallback behavior in the near future.

shared/search/providers.ts
efritz: Ah I see!

Pushed an update. Merging now, but could you check that one last time for correctness?

shared/search/promises.ts
creachadair: I'd simply like to avoid unnecessary premature abstraction. With this simplification, you wouldn't arguably even need a test.

PR #241, opened 2020-02-18
Add a basic code intel extension for Protocol Buffer (.proto) files.

creachadair:
Addresses part of https://github.com/sourcegraph/sourcegraph/issues/3688.

Our default ctags configuration already enables Protobuf, and covers the top-level productions.


efritz (approved):
If you publish the extension, publish **only** the new one as we need to smoke test the others a bit as there have been significant changes.

PR #245, opened 2020-02-19
Support product versions like 3.13.0-rc.1.

efritz:
Version comparison fails against rc tagged versions.

cc @mrnugget You may have the same bug after porting this code to src-cli.

### DIFF
--- a/shared/util/versions.ts
+++ b/shared/util/versions.ts
@@ -34,7 +34,7 @@ export function compareVersion({
     }
 
     // Split any -rc.x tags from the end of the product version
-    const semanticVersion = productVersion.split('-')[0]
+    const semanticVersion = productVersion.split(/[^\d.]/)[0]
     if (semver.valid(semanticVersion)) {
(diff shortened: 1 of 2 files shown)


shared/util/versions.ts
```diff
@@ -33,8 +33,10 @@ export function compareVersion({
         return enableForDev
     }
 
-    if (semver.valid(productVersion)) {
-        return semver.satisfies(productVersion, `>=${minimumVersion}`)
+    // Split any -rc.x tags from the end of the product version
+    const semanticVersion = productVersion.split('-')[0]
```
creachadair: We might not use it, but `1.2.3+blah` is also grammatically legal (as is `1.2.3-foo+blah`).

creachadair (approved):

shared/util/versions.ts
efritz: Updated to be a little more general.

PR #246, opened 2020-02-19
Lint pass

efritz:


creachadair:
Thanks for doing this. Are there any lint hooks I should have set up for when I'm touching this code?

efritz:
No, I'm waiting for a release for our base eslint config to become more sane. Right now we're not applying all the rules we want/too many rules. The sg/sg repo doesn't have this for some reason because they're not skipping the react rules.

You _can_ run `yarn tslint/eslint` but there's a lot of noise right now.

Once that dependency package is released to npm I'll let you know.

PR #251, opened 2020-02-21
Update docs and manifests

efritz:
This gets the docs and settings ready for the first publication of extensions after the great refactor.

- Updated readme for each extension. This includes a section on basic code intel, its limitations, and how to configure lsif. It also gives some instructions on how to disable unindexed search, which is bad for some current prospective customers that use a monorepo.
- Pruned menus and actions that were effectively dead (disable intel, show feedback, etc).
- Pruned settings that are no longer used from TS extension.


extensions/go/README.md
```diff
@@ -2,13 +2,65 @@
...
+
+-   "Go to definition" on a token goes to the definition found by [universal-ctags](https://github.com/universal-ctags/ctags), a cross-language parsing suite.
+-   "Find references" on a token finds all instances of token (with the same case) in the current repository and other repositories.
+
+These heuristics work well for tokens with unique names, such as `render_to_view` or `TLSConfig`. They do not work well for ambiguous tokens, such as `open` or `file`.
+
+### Large repositories
+
+Heuristic code intelligence will perform a search query in the commit you are viewing. This may cause performance issues if the commit is not indexed and the repository is large. After a timeout period with no results, an index-only search will be performed. This type of query may return results for a commit other than the one you are currently viewing. The default timeout period is five seconds, but can be lowered by adding the following to your Sourcegraph global settings:
```
creachadair:
```suggestion
Basic code intelligence will perform a search query in the commit you are viewing. This may cause performance issues if the commit is not indexed and the repository is large. After a timeout period with no results, an index-only search will be performed. This type of query may return results for a commit other than the one you are currently viewing. The default timeout period is five seconds, but can be lowered by adding the following to your Sourcegraph global settings (units are milliseconds):
```

Since "basic" is the term we defined above, we should probably be consistent here.

Re: Units: I wrote milliseconds but whatever is true here.

extensions/go/README.md
```diff
@@ -2,13 +2,65 @@
...
+
+    ```json
+    "basicCodeIntel.unindexedSearchTimeout": 1000
+    ````
+
+For organizations that organize code in a monorepo, it may never be useful to perform an un-indexed search. Index-only search queries can be forced by adding the following to your Sourcgraph global settings.
```
creachadair:
```suggestion
For organizations that organize code in a monorepo, it may never be useful to perform an un-indexed search. To force only indexed search queries, add the following to your Sourcgraph global settings.
```

Inspired by https://about.sourcegraph.com/handbook/communication/style_guide#avoid-jargon-and-acronyms, https://about.sourcegraph.com/handbook/communication/style_guide#clarity

extensions/go/README.md
```diff
@@ -2,13 +2,65 @@
...
+
+To enable [LSIF support](https://docs.sourcegraph.com/user/code_intelligence/lsif), add these to your Sourcegraph global settings:
+
+```json
+  "codeIntel.lsif": true
+```
+
+## Language server
+
+This extension is configured to talk to a language server over WebSockets. On Sourcegraph.com, this extension is already configured. If you are running a
```
creachadair:
```suggestion
This extension communicates with a language server over WebSockets. On Sourcegraph.com, this extension is already configured. If you are running a
```

Unless we can configure it to communicate some other way.

extensions/go/README.md
```diff
@@ -115,32 +167,7 @@ Now visit a Go file and you should see code intelligence!
...
-          port: 6060
-          targetPort: debug
-        - name: lsp
-          port: 443
-          targetPort: lsp
-    selector:
-        app: lang-go
-    type: LoadBalancer
-```
+If you want to deploy the language server with Kubernetes, your deployment should look like this:
```
creachadair:
```suggestion
To deploy the language server with Kubernetes, use a deployment like this:
```

extensions/go/README.md
```diff
@@ -228,7 +249,50 @@ spec:
...
+          port: 443
+          targetPort: lsp
+    selector:
+        app: lang-go
+    type: LoadBalancer
+```
+
+#### TLS
+
+TLS is optional but recommended for production deployments. It is used if `TLS_KEY` and `TLS_CERT` environment variables are set.
```
creachadair:
```suggestion
To enable TLS, set the `TLS_KEY` and `TLS_CERT` environment variables. TLS optional but **strongly recommended** for production deployments.
```

Is `TLS_CERT` a path or an encoded certificate?

extensions/go/README.md
```diff
@@ -228,7 +249,50 @@ spec:
...
+
+#### Scaling out by increasing the replica count
+
+You can run multiple instances of the go-langserver and distribute connections between them in Kubernetes by setting `spec.replicas` in the deployment YAML:
```
creachadair:
```suggestion
To run multiple instances of the go-langserver and distribute connections between them in Kubernetes, set `spec.replicas` in the deployment YAML:
```

extensions/go/src/settings.ts
```diff
@@ -23,11 +23,11 @@ export interface Settings {
      */
     'basicCodeIntel.unindexedSearchTimeout'?: number
     /**
-     * The WebSocket language server to connect to
+     * The address of the WebSocket language server to connect to.
```
creachadair: Maybe give an example for format? `wss://www.example.com/socketserver`  or `host:port` or whatever.

extensions/go/README.md
creachadair: (If you accept these changes they'll also need to be propagated back into the template of course)

extensions/template/README.md
```diff
@@ -1,53 +1,60 @@
...
+1. Visit any file on your code host and hover over a token to see a description of the token, a **Go to definition** action (if available), and a **Find references** action.
 
-![screenshot of using LANGNAME on GitHub](https://user-images.githubusercontent.com/1976/50882271-0c5f3180-1399-11e9-9697-e4e4fa4e29e9.png)
+![screenshot of using LANGID on GitHub](https://user-images.githubusercontent.com/1976/50882271-0c5f3180-1399-11e9-9697-e4e4fa4e29e9.png)
 
-### Limitations
+## Basic code intelligence
 
-Because this extension uses text-based heuristics, its definition and reference results are not precise:
+This extension comes with built-in code intelligence provided by [search-based heuristics](https://docs.sourcegraph.com/user/code_intelligence/basic_code_intelligence). Because this extension uses text-based heuristics, its definition and reference results may not be precise:
```
creachadair: I think "are not" was better: They may be _correct_ (for a given query) but they are not _precise_.

extensions/go/README.md
efritz: Updated in all three places.

extensions/go/README.md
efritz: Updated in all three places.

extensions/go/README.md
efritz: Updated in both LSP extensions.

extensions/go/README.md
efritz: Updated in both LSP extensions.

extensions/template/README.md
efritz: Updated in all three places.

extensions/go/README.md
efritz: Updated.

extensions/go/src/settings.ts
efritz: Updated in both configs, regenerated settings files.

extensions/go/README.md
efritz: Looks like an encoded certificate based on the config in https://github.com/sourcegraph/deploy-sourcegraph-dot-com.

creachadair (approved):

PR #254, opened 2020-02-21
Publish extensions sequentially.

efritz:
It was a mistake to do them all in parallel...


creachadair (approved):
This seems good for reproducibility, but did this cause a real problem?

efritz:
On my local machine, yes. I'd like to be able to use my cursor, even when running a command.

creachadair:
Unbounded concurrency will get you every time.

PR #263, opened 2020-02-26
Fix repo urls in package.json files.

efritz:


creachadair (approved):

extensions/go/package.json
```diff
@@ -9,7 +9,7 @@
   "license": "MIT",
   "repository": {
     "type": "git",
-    "url": "https://github.com/sourcegraph/code-intel-extensions/extensions/go"
+    "url": "https://github.com/sourcegraph/code-intel-extensions/tree/master/extensions/go"
```
felixfbecker: The `url` field is supposed to be a git-cloneable URL, so this is not valid unfortunately. You can specify the subdirectory with the `directory` field (the TS langserver also uses that for cross-repo code intel). See https://docs.npmjs.com/files/package.json#repository

> The URL should be a publicly available (perhaps read-only) url that can be handed directly to a VCS program without any modification. It should not be a url to an html project page that you put in your browser. It’s for computers.

> If the package.json for your package is not in the root directory (for example if it is part of a monorepo), you can specify the directory in which it lives:

PR #264, opened 2020-02-27
Fix document selector scope

lguychard:
Rel https://github.com/sourcegraph/sourcegraph/issues/8602

### DIFF
--- a/shared/lsp/registration.ts
+++ b/shared/lsp/registration.ts
@@ -607,9 +607,11 @@ export function scopeDocumentSelectorToRoot(
         )
         .map(filter => ({
             ...filter,
-            // TODO filter.pattern needs to be run resolved relative to server root URI before mounting on clientRootUri
-            // Root URIs are of the form git://repo?commit-sha, so new URL(pattern, URI) would strip the query string,
-            // resulting in a pattern that matches any commit ID.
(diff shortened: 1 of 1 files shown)


shared/lsp/registration.ts
```diff
@@ -608,6 +608,8 @@ export function scopeDocumentSelectorToRoot(
         .map(filter => ({
             ...filter,
             // TODO filter.pattern needs to be run resolved relative to server root URI before mounting on clientRootUri
-            pattern: new URL(filter.pattern ?? '**', clientRootUri).href,
+            // Root URIs are of the form git://repo?commit-sha, so new URL(pattern, URI) would strip the query string,
+            // resulting in a pattern that matches any commit ID.
+            pattern: `${clientRootUri.href}#${filter.pattern ?? '**/**'}`,
```
lguychard: We end up using minimatch to match this, and while `**` _should_ suffice, for some reason it didn't. We might also want to tweak our matching logic in https://sourcegraph.com/github.com/sourcegraph/sourcegraph@9ae2eae3dae2fd6f330b7a926ec4d40db45a5bfe/-/blob/shared/src/api/client/types/textDocument.ts#L64-88...

efritz (approved):

PR #266, opened 2020-02-27
Fix performing a search on illegal ranges.

efritz:
This stops, for example, hovering over `\t` in go code.

### DIFF
--- a/shared/search/tokens.ts
+++ b/shared/search/tokens.ts
@@ -53,6 +53,10 @@ export function findSearchToken({
         }
     }
 
+    if (start >= end) {
+        return undefined
+    }
+
     const searchToken = line.substring(start, end)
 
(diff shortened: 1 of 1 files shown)


shared/search/tokens.ts
```diff
@@ -53,6 +53,10 @@ export function findSearchToken({
         }
     }
 
+    if (start >= end) {
```
creachadair: Isn't `start === end` a legal representation of an empty range? (Or do we not handle empty ranges?)

efritz: Empty ranges (as in `[x, x)`) can't be hovered over, so that's fine!

I'd have to check to see if this a half-open interval, though. Either way: it was the previous behavior, and we're missing at most single-letter basic code intel, which isn't going to give us much precision even if we perform it.

PR #267, opened 2020-02-28
Do not register multiple cached lsp connections for the same root uri.

efritz:


efritz:
This should **actually** close https://github.com/sourcegraph/sourcegraph/issues/8602.

creachadair (approved):

PR #269, opened 2020-02-28
Refactor LSP registration

efritz:
Remove support for multiple workspaces (we only have two language servers and neither of them support this). It's a lot of registration code that's just dead weight.

Ensure that providers are unsubscribed when we kill the underlying LSP connection. This was _actually_ (third try should finally do it) the root cause of https://github.com/sourcegraph/sourcegraph/issues/8602.

Reviewing by-commit might be easiest.

### DIFF
--- a/shared/lsp/registration.ts
+++ b/shared/lsp/registration.ts
@@ -386,7 +386,6 @@ export async function register({
         try {
             return await fn(connection)
         } finally {
-            connectionsByRootUri.delete(workspaceFolder.href)
             connection.unsubscribe()
         }
     }
--- a/shared/lsp/registration.ts
+++ b/shared/lsp/registration.ts
@@ -1,5 +1,4 @@
 import { differenceBy, identity } from 'lodash'
-import * as path from 'path'
 import { EMPTY, from, Observable, Subscription, Unsubscribable } from 'rxjs'
 import { map, scan, startWith } from 'rxjs/operators'
 import * as sourcegraph from 'sourcegraph'
@@ -9,10 +8,7 @@ import { Logger, LogLevel } from '../logging'
 import { ProviderWrapper } from '../providers'
 import { LSPClient } from './client'
 import { LSPConnection } from './connection'
-import {
-    convertDiagnosticToDecoration,
-    toLSPWorkspaceFolder,
-} from './conversion'
+import { convertDiagnosticToDecoration } from './conversion'
 import {
     definitionFeature,
     DefinitionFeatureOptions,
@@ -77,7 +73,6 @@ const clientCapabilities = {
 export interface RegisterOptions {
     progressSuffix?: string
     sourcegraph: SourcegraphAPI
-    supportsWorkspaceFolders?: boolean
     clientToServerURI?: (uri: URL) => URL
     serverToClientURI?: (uri: URL) => URL
     logger?: Logger
@@ -95,7 +90,6 @@ export async function register({
     serverToClientURI = identity,
     logger = console,
     progressSuffix = '',
-    supportsWorkspaceFolders,
     transport: createConnection,
     documentSelector,
     initializationOptions,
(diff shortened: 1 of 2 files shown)


shared/lsp/registration.ts
```diff
@@ -363,203 +361,106 @@ export async function register({
         fn: (connection: LSPConnection) => Promise<R>
     ) => Promise<R>
 
-    if (supportsWorkspaceFolders) {
-        const connection = await connect({
-            clientRootUri: null,
+    // Supports only one workspace root
+    // TODO this should store a refcount to avoid closing connections other consumers have a reference to
```
creachadair: Makes sense, but _are_ there other consumers? Doesn't our LS run inside the container network where it won't be reachable?

shared/lsp/registration.ts
```diff
@@ -363,203 +361,106 @@ export async function register({
...
-                        {
-                            event: {
-                                added: [],
-                                removed: [tempWorkspaceFolder],
-                            },
-                        }
+        try {
+            return await fn(connection)
+        } finally {
+            connectionsByRootUri.delete(workspaceFolder.href)
```
creachadair: I think I'm missing something. Reaching 373 appears to mean `connectionsByRootUri` did not contain `workspaceFolder.href`, or else we would have exited at 371. Should we have added it before 386?

shared/lsp/registration.ts
```diff
@@ -363,203 +361,106 @@ export async function register({
...
+                            capabilities: clientCapabilities,
+                            workspaceFolders: null,
+                            initializationOptions,
+                        },
+                        registerProviders: true,
+                    })
+                    subscriptions.add(connection)
+                    return connection
+                } catch (err) {
+                    logger.error('Error creating connection', err)
```
creachadair: Can we also log the root URI, so we can tell what the workspace is that failed?

shared/lsp/registration.ts
```diff
@@ -363,203 +361,106 @@ export async function register({
...
-                        return after
-                    })
-                )
-                .subscribe()
-        )
-        addRoots(sourcegraph.workspace.roots)
-        await Promise.all(connectionsByRootUri.values())
-    }
+                            } catch (err) {
+                                logger.error('Error disposing connection', err)
```
creachadair: As 414.

creachadair (commented):
Just a few small things.

shared/lsp/registration.ts
efritz: This isn't related to the PR. This also runs in the browser so it's only a single user. I'm not sure how useful it would be to do this, but xrefs providers can temporarily hold a connection that may be closed here.

shared/lsp/registration.ts
efritz: This can throw async. We put the _promise_ in the map and then if the promise fails during init we'll remove ourselves.

shared/lsp/registration.ts
efritz: This is also unrelated to the PR. We can do this but in a separate effort.

shared/lsp/registration.ts
efritz: This is also unrelated to the PR. We can do this but in a separate effort.

shared/lsp/registration.ts
creachadair: Right—and during `addRoots` it works that way. But here we aren't adding any mass to the map, and we've already ruled out that it was present before.

What I'm not following is that in the case where we _did_ find the promise in the map, we didn't remove it when we're done; but here (where it's not) we do.  I realize the promise may have failed between creation and now, but the registration in the map happens sooner as far as I can see.

efritz: Oops, sorry. I was thinking this was another similar line in the file. You're right that this is unnecessary.

creachadair (approved):

shared/lsp/registration.ts
felixfbecker: > xrefs providers can temporarily hold a connection that may be closed here.

That is indeed the issue.

PR #280, opened 2020-03-02
Fix repo URL values in package definitions

efritz:


extensions/go/package.json
```diff
@@ -9,7 +9,8 @@
   "license": "MIT",
   "repository": {
     "type": "git",
-    "url": "https://github.com/sourcegraph/code-intel-extensions/tree/master/extensions/go"
+    "url": "github:sourcegraph/code-intel-extensions",
```
felixfbecker: I wasn't aware this `github:` shorthand was also legal and had to check whether the TS langserver also supported it, turns out it does not :/

https://github.com/sourcegraph/sourcegraph-typescript/blob/master/src/dependencies.ts#L228-L231

We can either stick with the normal https:// URLs, the non-prefixed shorthand and/or fix the langserver

felixfbecker: https://github.com/sourcegraph/sourcegraph-typescript/pull/336

efritz: Approved.

PR #281, opened 2020-03-02
Update test URL

lguychard:


efritz:
This test fails now. Was there an issue with the original change?

efritz:
Thanks!

PR #290, opened 2020-03-05
Remove cached version of queryGraphQL

efritz:
We should move all caching to the provider level, not on individual requests. The only place we were caching previously was for LSIF results, but the pages for references can be large.

We aren't able to cache search results due to timeouts that were affecting large monorepo customers, so the previous caching strategy was already lacking.

### DIFF
--- a/shared/lsif/graphql.ts
+++ b/shared/lsif/graphql.ts
@@ -12,7 +12,7 @@ import { LocationConnectionNode, nodeToLocation } from './conversion'
  * relatively low unless it's a VERY popular library and LSIF data is
  * ubiquitous (which is our goal).
  */
-const MAX_REFERENCE_PAGE_REQUESTS = 20
+const MAX_REFERENCE_PAGE_REQUESTS = 5
 
 /**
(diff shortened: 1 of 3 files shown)


shared/lsif/graphql.ts
```diff
@@ -12,7 +12,7 @@ import { LocationConnectionNode, nodeToLocation } from './conversion'
  * relatively low unless it's a VERY popular library and LSIF data is
  * ubiquitous (which is our goal).
  */
-const MAX_REFERENCE_PAGE_REQUESTS = 20
+const MAX_REFERENCE_PAGE_REQUESTS = 5
```
creachadair: No objection, but why this change? Does removal of caching make the previous limit bad?

creachadair (approved):

shared/lsif/graphql.ts
efritz: No, this is kind of a drive-by.

efritz: Actually I'll revert this line.

PR #293, opened 2020-03-05
Add a very small LRU cache to search providers

efritz:
Add a small cache when fetching file contents, as we currently do this for every file that contains a definition (once per hover). If the definition is in the same file or if the definition is in a file we've recently seen, we can save that fetch.

Add another small cache to the search provider definition result. This isn't meant to help looking at the same definition twice, but _will_ reduce the number of queries. The hover provider currently calls definition to get the first value from which to pull the hover text. We make the same 2-4 queries per hover action in the API. We can cut the second redundant query (or pair of queries) entirely.


shared/search/providers.ts
```diff
@@ -16,6 +17,12 @@ import { definitionQuery, referencesQuery } from './queries'
 import { BasicCodeIntelligenceSettings } from './settings'
 import { findSearchToken } from './tokens'
 
+/** The number of elements in the document LRU cache. */
+const DOCUMENT_CACHE_SIZE = 50
```
creachadair: Does the cache allow us to plug in an element size metric? With file contents, you can easily blow your RAM budget without exceeding your element budget. Where practical, I like to use file size as the cache metric instead of treating each file as an atom.

But a lot of cache implementations don't let you customize the metric, so I wouldn't block on this. Instead what I'd suggest in that case is to set a size cap above which we'll refuse to cache the file contents.

efritz: It's got a length function: https://www.npmjs.com/package/lru-cache

efritz: It's problematic though because we're caching promises to elements we don't know the size of, not the elements itself.

creachadair: Yuck, that's both good and bad. The promises aren't very big, but the data might be. Maybe we should await on the promise when we (try) to do a cache insert, and only insert if the size is sane? 

Otherwise, I worry we are blowing a giant hole in our RAM budget. Not that we really have a budget, I guess.

creachadair: Now that I think about it, I think it should be safe to await on it at the point of insertion: The promise we grab on either arm of that conditional is because the caller wants the file content, so blocking for that content should not stall the rest of the transaction any additional amount. Then the cache's existing max-size handling will just work.

efritz: I'm not sure I understand. What happens when the order is:

1. user requests file content, isn't cached
1. goes to fetch file content, doesn't return immediately
1. second request for file content (still isn't cached), starts new request
1. first request finishes, gets put in cache
1. second request finishes, overwrites cache

efritz: Unless we proactively put something in the cache so the same promise resolves for all those that grab it, but if it turns out to be an insane size we can immediately delete it. Then it's only live for the time of the handful of requests that were active at the time.

creachadair: That might be the least-ugly workaround that doesn't require a lot of extra code. I'm reluctant to do something really complicated, because I think in practice this won't usually be a problem. 

If we're willing to let the size burst temporarily, I suppose we could get the same effect by setting `maxAge` and `updateAgeOnGet`, and then maybe we don't need to change anything else about how we use the cache.

efritz: This will also only affect one user as it’s running in a web worker context. We solved this problem in lsif-server but it requires hand writing the cache so that we can effectively track the sizes of the resolved values.

creachadair: All right. I guess maybe the most practical thing to do in this case is to just cross our fingers and hope it doesn't happen too much. 😞

creachadair (approved):

shared/search/providers.ts
efritz: I've removed the document cache until we can figure out what to do with it. The real change was stopping definition/hover from making the same queries at once, this just seemed like a nice drive-by.

PR #303, opened 2020-03-06
Scope DocumentSelector with baseUri

### LINKED ISSUE #8602: Duplicate type signatures/docs in hover tooltips
Sourcegraph.com/GitHub (Chrome extension)

I've seen this multiple times today, always on GitHub. No idea how to repro, and after refreshing it disappears. The first time I saw them, there were 3x, not just 2x (as you see here).

image

image

felixfbecker:
This is a proper fix for both https://github.com/sourcegraph/sourcegraph/issues/8602 and broken codeintel for root files ([Slack thread](https://sourcegraph.slack.com/archives/CHXHX7XAS/p1583444612193300?thread_ts=1583444599.193000&cid=CHXHX7XAS)).
Codeintel extensions need to be able to scope providers to workspace roots. Doing so in `pattern` does not work, because `minimatch` is not built for working with URIs (in particular it broke for files in the root).

Instead, uses the new `baseUri` introduced in https://github.com/sourcegraph/sourcegraph/pull/8858.

Closes https://github.com/sourcegraph/code-intel-extensions/pull/295.
Fixes sourcegraph/sourcegraph#8602.

### DIFF
--- a/.buildkite/yarn-run.sh
+++ b/.buildkite/yarn-run.sh
@@ -6,6 +6,9 @@ cd $(dirname "${BASH_SOURCE[0]}")/..
 # mutex is necessary since CI runs various yarn installs in parallel
 yarn --mutex network --frozen-lockfile --network-timeout 60000
 
+# Need to build for project references
+yarn build-ts
+
 for cmd in "$@"; do
     yarn -s --cwd "${CWD:-.}" run ${cmd}
 done
--- a/dev/scripts/tsconfig.json
+++ b/dev/scripts/tsconfig.json
@@ -1,5 +1,6 @@
 {
   "extends": "../../tsconfig.json",
+  "references": [{ "path": "../../shared" }],
   "compilerOptions": {
     "outDir": "dist"
   },
--- a/extensions/typescript/src/xrefs.ts
+++ b/extensions/typescript/src/xrefs.ts
@@ -157,7 +157,6 @@ async function findExternalRefsInDependent(
         return []
     }
     const rootUri = new URL(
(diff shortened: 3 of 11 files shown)


efritz:
Am I correct that publishing this will break all current instances?

felixfbecker:
No, it should not. Current instances will ignore the `baseUri` property. They will just continue to see the double hovers unfortunately until they upgrade to 3.14.

efritz:
But we’re no longer setting pattern so won’t no providers be registered correctly on old instances?

felixfbecker:
But the `pattern` was just `**` before (i.e. all files), and we're still setting `language`. If the language server provided a pattern we're still passing that through (just "unscoped"), but none of our language servers use that afaik.

felixfbecker:
(please sanity check what I'm saying, it's late 😅 )

efritz:
Go to bed I’ll test it later tonight and merge :)

efritz:
I ran this and it works neither with DotCom (#8858 has been merged by now and the site version has a rev after that merge commit) nor with a private instance before the merge.

efritz (changes requested):
Needs more manual testing.

felixfbecker:
I'm very confused. I thought the _Go_ issue were fixed by https://github.com/sourcegraph/code-intel-extensions/pull/292 (the revert), and we only had double hovers now?

felixfbecker:
I found the issue, fix in https://github.com/sourcegraph/sourcegraph/pull/8877. The issue was in the extension host, i.e. the changes in this PR are correct.

Verified that codeintel works:

image

efritz (approved):

PR #304, opened 2020-03-06
Add search provider tests

efritz:
Add unit tests to search providers. Require refactoring of API functions, which are now in a mockable class.


shared/util/graphql.ts
```diff
@@ -49,3 +49,8 @@ function aggregateErrors(errors: Error[]): Error {
         errors,
     })
 }
+
+export const gql = (
+    template: TemplateStringsArray,
+    ...substitutions: any[]
+): string => String.raw(template, ...substitutions)
```
felixfbecker: FYI this is not the same as `tagged-template-noop` and actually wrong :(
`String.raw` prevents any escape sequences in the string, which is not the intention (the intention is to just be a noop, like `tagged-template-noop`). I wrote this in the main repo, realized later that it was wrong, but changing would mean all the queries need to be tested again. In all other repos I used `tagged-template-noop` which I only discovered after.

efritz: I'm having issues including that due to a tsconfig issue. Same with lru-cache. When I run the tests when it's imported I get

`TypeError: lru_cache_1.default is not a constructor`

what's the resolution for these kinds of issues?

efritz: Solved it!

felixfbecker: Glad you did. For visibility: The resolution is `esModuleInterop: true` in tsconfig.json, and making sure the right import style is used: `import foo from 'bar'` imports the _default export_, `import * as foo from 'bar'` import _everything_ as a module record object (which can never be a function).

PR #313, opened 2020-03-09
Remove btoa hack

efritz:


shared/badges.ts
```diff
@@ -1,5 +1,4 @@
-const g: any = global
-g.btoa = g.btoa || ((s: string) => Buffer.from(s, 'binary').toString('base64'))
```
efritz: eeew

shared/badges.ts
```diff
@@ -1,5 +1,4 @@
...
+import btoa from 'btoa'
```
felixfbecker: 🤔 The package page says that it uses `Buffer` under the hood, which I think means that this pulls in a (probably large) `Buffer` polyfill into the extension bundle. A pure-JS Base64 implementation may be better suited, e.g. https://www.npmjs.com/package/js-base64

felixfbecker (approved):
✨

PR #319, opened 2020-03-10
Remove LSP access token check

### LINKED ISSUE #8915: Not signed in users only get basic code intel
This affects all users of sourcegraph.com, native integration on gitlab.com, and any user of the browser extension on any code host. Those users get inferior codeintel results in their very first experience with the product, which is not ideal. Basic code intel also often returns multiple results, which means go-to-definition doesn't stay on the code host, but links to sourcegraph.com. Users may assume that is always the case, and it's unfortunate because we **have** stay-on-code-host implemented as a feature.

The code intel extensions currently bail out on precise codeintel if no access token is in the settings (it is dynamically created, which can only be done if the user is signed in). But for public instances like sourcegraph.com, there is no technical reason for this, and the language server should work fine without an access token.

Context: https://sourcegraph.slack.com/archives/CHXHX7XAS/p1583836104283400

efritz:
Fixes https://github.com/sourcegraph/sourcegraph/issues/8915.

### DIFF
--- a/shared/activate.ts
+++ b/shared/activate.ts
@@ -89,6 +89,7 @@ const DUMMY_CTX = {
  * @param selector The document selector for which this extension is active.
  * @param languageSpec The language spec used to provide search-based code intelligence.
  * @param lspFactory An optional factory that registers an LSP client.
+ * @param logger An optional logger instance.
  */
 export async function activateCodeIntel(
     ctx: sourcegraph.ExtensionContext = DUMMY_CTX,
@@ -110,6 +111,7 @@ export async function activateCodeIntel(
  * @param ctx  The extension context.
  * @param wrapper The provider wrapper.
  * @param lspFactory An optional factory that registers an LSP client.
+ * @param logger An optional logger instance.
  */
 export async function tryInitLSP(
     ctx: sourcegraph.ExtensionContext,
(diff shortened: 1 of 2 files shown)


shared/lsp/auth.ts
```diff
@@ -20,18 +20,18 @@ export async function getOrCreateAccessToken(
...
     if (accessToken) {
         return accessToken
     }
     if (!accessTokenPromise) {
-        accessTokenPromise = (async (): Promise<string | undefined> => {
+        accessTokenPromise = (async (): Promise<string> => {
             const userId = await api.getUser()
             if (!userId) {
-                return undefined
+                return ''
```
felixfbecker: This seems weird - in TS, if something is typed as a `string` with a given semantic "access token" I would never expect it to be an empty string, because an empty string is not an access token. If the scenario is possible that there is no access token, that would be indicated and made type-checkable by making it a union `string | undefined` like before. If the thought behind this is to concatenate it into a string, I'd use `?? ''` at the place where it's concatenated

felixfbecker:
Did you test this with Go and TypeScript?

shared/lsp/auth.ts
felixfbecker: I think this is especially important because we need to make sure the consumers don't create a URI like `https://@sourcegraph.com/...` but handle the case where it's not defined, to also omit the `@`

felixfbecker:
Why the close? 😢

efritz:
I reset the branch and it auto-closed :/

efritz:
https://github.com/sourcegraph/code-intel-extensions/pull/320

PR #320, opened 2020-03-10
Remove LSP access token check

efritz:


felixfbecker:
Did you test with TS and Go?

efritz:
Yes.

felixfbecker (approved):

PR #323, opened 2020-03-11
Update instructions for publishing extensions.

creachadair:
The text refers to instructions for publishing, but they were missing.
Add them, update the reference, and fix a few lexical nits.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,9 @@ This repository contains the code for the [Sourcegraph extensions that provide c
 
 ## Generating & publishing extensions
 
-Extensions are generated and published from the [master branch](https://buildkite.com/sourcegraph/code-intel-extensions/builds?branch=master).
+Typically you do not need to explicitly generate and publish extensions. By default, extensions are generated and published in BuildKite from the [master branch](https://buildkite.com/sourcegraph/code-intel-extensions/builds?branch=master).
+
+If you need to manually generate/publish extensions, do so as follows:
 
 Generate:
 
--- a/README.md
+++ b/README.md
@@ -15,28 +15,39 @@ This repository contains the code for the [Sourcegraph extensions that provide c
 ## Development
 
 1. Run `yarn`
-1. Run `yarn --cwd extensions/{go,typescript,template} run serve` (pick one, `template` includes all others)
-1. Open up your Sourcegraph settings https://sourcegraph.com/users/you/settings and disable the language extensions you're developing:
+2. Run `yarn --cwd extensions/{go,typescript,template} run serve` (pick one, `template` includes all others)
+3. Open up your Sourcegraph settings https://sourcegraph.com/users/you/settings and disable the language extensions you're developing:
 
-```json
-{
-  ...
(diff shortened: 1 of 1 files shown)


creachadair:
I'm not 100% sure I got all the corner cases correct. Advice welcome.

README.md
```diff
@@ -15,28 +15,44 @@ This repository contains the code for the [Sourcegraph extensions that provide c
...
-1. Make sure there is a mapping entry for the `languageID` in https://github.com/sourcegraph/sourcegraph/blob/master/shared/src/languages.ts#L40
-1. Generate and publish the extension as described in the previous section.
+2. (optional, to enable jump to definition) Ensure the language is present in the command line arguments to universal-ctags https://github.com/sourcegraph/sourcegraph/blob/21efc6844838e773b9a8f4a7ba1d5628e8076984/cmd/symbols/internal/pkg/ctags/parser.go#L71
+3. Make sure there is a mapping entry for the `languageID` in https://github.com/sourcegraph/sourcegraph/blob/master/shared/src/languages.ts#L40
+4. Generate and publish the extension as described below.
 
-## Publishing extensions
+## Generating & publishing extensions
 
 Extensions are generated and published from the [master branch](https://buildkite.com/sourcegraph/code-intel-extensions/builds?branch=master).
```
efritz: Should we have explicit publishing instructions if it's done from the master branch in CI now?

README.md
```diff
@@ -15,28 +15,44 @@ This repository contains the code for the [Sourcegraph extensions that provide c
...
+    ```
 
 4. [Sideload the extension](https://docs.sourcegraph.com/extensions/authoring/local_development) (hit OK on the alert to accept the default URL http://localhost:1234) on your Sourcegraph instance and refresh the page. Make sure you don't see two of the same language extension in the **Ext** menu.
 
 ## Adding a language extension
 
 1. Add an entry to the [`shared/language-specs`](shared/language-specs) directory. For languages with a trivial configuration add a new entry (in alphabetical order) to [`languages.ts`](shared/language-specs/languages.ts). Otherwise, add additional files following the existing examples.
-1. (optional, to enable jump to definition) Ensure the language is present in the command line arguments to universal-ctags https://github.com/sourcegraph/sourcegraph/blob/21efc6844838e773b9a8f4a7ba1d5628e8076984/cmd/symbols/internal/pkg/ctags/parser.go#L71
-1. Make sure there is a mapping entry for the `languageID` in https://github.com/sourcegraph/sourcegraph/blob/master/shared/src/languages.ts#L40
-1. Generate and publish the extension as described in the previous section.
```
efritz: Oof good catch.

README.md
creachadair: I still think it's worthwhile to know how to do it manually. But maybe I should make it more explicit that the default is "just sit back and let CI do it"?

creachadair: I added some text to spell that out, PTAL.

efritz (approved):

PR #328, opened 2020-03-12
Refresh websocket connection after idle

### LINKED ISSUE #8963: Code intel stops working after a couple of minutes on idle page
- **Sourcegraph version:** Master
- **Platform information:** Chrome, Mac OS X

This may be LSIF-specific, I'm not sure.

#### Steps to reproduce:

1. Visit [a Go file in our repo](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/cmd/frontend/graphqlbackend/search.go#L123:9)
2. Hover on `NewSearchImplementer` and see results (the specifics doesn't really matter, choose anything to hover on).
3. Go do something else for 5 minutes
4. No pop up hovers for `NewSearchImplementer` or anything else. Basically, no code intel.

I didn't see anything come up in the console. It works again after a page refresh.

efritz:
Fixes https://github.com/sourcegraph/sourcegraph/issues/8963. For additional context see [this slack thread](https://sourcegraph.slack.com/archives/CHXHX7XAS/p1583990106335100).

When a WebSocket connection closes on idle, it unsubscribes all the features attached to that connection. For root connections, this manifests as not having any providers subscribed after leaving the UI open for a couple of minutes.

This PR will detect a root connection's close event and will re-register a connection for that same root if it's still in the cache map. This won't re-open connections that have been closed due to navigation. Re-connecting will also re-register the providers that were attached to a previous connection.


shared/lsp/registration.ts
```diff
@@ -389,33 +389,47 @@ export async function register({
         }
     }
 
+    function addRoot(root: sourcegraph.WorkspaceRoot): void {
+        const connectionPromise = (async () => {
+            try {
+                const serverRootUri = clientToServerURI(
+                    new URL(root.uri.toString())
```
felixfbecker: Isn't `root.uri` already a `URL` object?
```suggestion
                    root.uri
```

shared/lsp/connection.ts
```diff
@@ -24,12 +24,12 @@ export interface LSPConnection extends sourcegraph.Unsubscribable {
 
 export const webSocketTransport = ({
     serverUrl,
-    logger,
     cancellationToken,
+    logger = console,
```
felixfbecker: We need to be careful with using `console` directly, it should be wrapped at least in a `RedactedLogger`. Logs may contain the root URI (in errors, objects or whatever) which contains the access token, and we had people accidentally post those logs to our public issue tracker. I'd recommend against a default value

shared/lsp/registration.ts
```diff
@@ -389,33 +389,47 @@ export async function register({
...
+                subscriptions.add(connection)
+
+                connection.closeEvent.subscribe(() => {
+                    if (connectionsByRootUri.has(root.uri.toString())) {
+                        logger.log(
+                            'Refreshing WebSocket connection to language server'
+                        )
+                        addRoot(root)
+                    }
+                })
```
felixfbecker:
```suggestion
                subscriptions.add(connection.closeEvent.subscribe(() => {
                    if (connectionsByRootUri.has(root.uri.toString())) {
                        logger.log(
                            'Refreshing WebSocket connection to language server'
                        )
                        addRoot(root)
                    }
                }))
```
I thought this would give an eslint warning... could you check if I need to fix the configuration?

shared/lsp/registration.ts
```diff
@@ -389,33 +389,47 @@ export async function register({
...
+                        rootUri: serverRootUri.href,
+                        capabilities: clientCapabilities,
+                        workspaceFolders: null,
+                        initializationOptions,
+                    },
+                    registerProviders: true,
+                })
+                subscriptions.add(connection)
+
+                connection.closeEvent.subscribe(() => {
```
lguychard: I may be missing something here, but won't this leak connections?

When a root is removed from the workspace (eg. when navigating away from a file), I would expect the connection to be unsubscribed from: https://sourcegraph.com/github.com/sourcegraph/code-intel-extensions/-/blob/shared/lsp/connection.ts#L86-89

This will call `socket.close()`, which in turn will cause the `close` event to fire on the socket, following which... We'll add another root.

shared/lsp/registration.ts
felixfbecker: Ah I think this is left over from a transition period where we had to support both strings and URL objects

felixfbecker (commented):
It's awesome to see connections being made more resilient!

shared/lsp/registration.ts
efritz: Removing a root should first remove it from the map, the. Unsubscribe. That’s why we check if there’s still an entry for it.

lguychard: I see (sorry, I missed "This won't re-open connections that have been closed due to navigation" in the PR description).

lguychard (approved):

shared/lsp/registration.ts
efritz: Updated. No lint.

shared/lsp/registration.ts
efritz: Updated in a few other places as well.

shared/lsp/connection.ts
efritz: Marked every use of console with TODO (will clean these up in a next commit).

felixfbecker (approved):

shared/lsp/registration.ts
felixfbecker: Sorry about that, I'll look into it

shared/lsp/connection.ts
felixfbecker: Sounds good, thanks. This rule may help to enable just in this repo (and disable with a comment in the place where it's used legitimately): https://eslint.org/docs/rules/no-console

efritz: Updated.

PR #331, opened 2020-03-19
Add options to include forks and archives in search

efritz:
https://github.com/sourcegraph/sourcegraph/pull/8739 changed the default behavior of search to exclude forked and archived repositories. This caused a regression in the code intel test suite as many of the test repositories (in the sourcegraph-testing org) are forked from a public source.

This PR adds options to enable searching in forked and archived repositories for code intel.

### DIFF
--- a/shared/search/providers.ts
+++ b/shared/search/providers.ts
@@ -373,8 +373,6 @@ async function searchReferences(
  * @param search The search function.
  * @param args The arguments to the search function.
  * @param negateRepoFilter Whether to look only inside or outside the given repo.
- * @param includeForks Whether to include forked repos in results.
- * @param includeArchived Whether to include archived repos in results.
  */
 export function searchWithFallback<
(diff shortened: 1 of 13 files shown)


shared/search/providers.ts
```diff
@@ -367,59 +372,96 @@ async function searchReferences(
...
-            query: indexedQuery,
-        })
+        return searchIndexed(search, args, negateRepoFilter)
     }
 
-    const timeout = getConfig<number>(
-        'basicCodeIntel.unindexedSearchTimeout',
-        5000
-    )
+    // searchIndexed/searchUnindexed modifies queryTerms so make
```
creachadair: Can we make it the responsibility of the callee not to diddle its arguments?

If I ever need to use these functions elsewhere I'm not going to remember this. 🙂

creachadair (approved):
Just one request, and not a blocker.

Issue #330, opened 2020-03-18
Hover fallback logic


efritz:
If there is no precise hover, we'll fall back to a search provider for hover text, which calls the search provider definition and scans the area around that text.

What should happen is that we ask for _the best definition available_ and use that to scan the text for a heuristic hover. The search-based hover provider should call the top level definition provider, not just its own.

efritz:
Fixed by https://github.com/sourcegraph/code-intel-extensions/pull/335.

PR #335, opened 2020-03-24
Feed a recursive top-level providers reference into search providers

efritz:
The search hover provider requests the most likely definition in order to extract comment text near that location. Currently search hover asks search definition, which isn't ideal in the case where an LSIF indexer has location data but no hover data. This PR passes a reference to the actively registered providers so the search hover can call "back to the top" of the definition stack.

See #330 for context.


shared/providers.ts
```diff
@@ -343,8 +363,13 @@ export function badgeValues<T extends object>(
 }
 
 /**
- * Converts an async generator provider into an observable provider. This also
- * memoizes the previous result as a workaround for #1321 (below).
+ * Converts an async generator provider into an observable provider. This
+ * also memoizes the previous result. This reduces the number of definition
+ * requests from two to one on each hover.
+ *
+ * The memoization was originally a workaround for #1321 (below), but should
```
creachadair: I didn't think of this when you made the original change. My only concern about  memoization is that it's potentially an unbounded memory leak. What's the dynamic extent of the wrapper? Is it coeval with the process or some smaller transaction?

Not a blocker here (the change was already made) but I wanted to bring it up because I'm concerned that it could be hard to fix later.

shared/search/providers.test.ts
```diff
@@ -437,6 +439,23 @@ describe('search providers', () => {
         })
     })
 
+    /** Create providers with the definition provider fed into itself. */
+    const recurProviders = (api: API): Providers => {
```
creachadair: Cute. This is roughly how parser combinators deal with recursive productions too. 🙂

creachadair (approved):
Tricky, but seems like a sensible approach.

PR #363, opened 2020-04-07
Search forks within fork

efritz:
If the user is currently browsing a fork, we should search in forks for definitions and references _in the same repository_, but not in remote repositories.

This was reported in https://github.com/sourcegraph/sourcegraph/issues/9634.

### DIFF
--- a/extensions/typescript/src/package.test.ts
+++ b/extensions/typescript/src/package.test.ts
@@ -7,7 +7,9 @@ describe('resolvePackageRepo', () => {
     it('resolves string repo', async () => {
         const api = new API()
         const mock = sinon.stub(api, 'resolveRepo')
-        mock.callsFake(repo => Promise.resolve(repo))
+        mock.callsFake(repo =>
+            Promise.resolve({ name: repo, isFork: false, isArchived: false })
+        )
         const name = await resolvePackageRepo('{"repository": "foo"}', api)
         assert.equal(name, 'foo')
         sinon.assert.calledWith(mock, 'foo')
@@ -16,7 +18,9 @@ describe('resolvePackageRepo', () => {
     it('resolves repos with url', async () => {
         const api = new API()
         const mock = sinon.stub(api, 'resolveRepo')
-        mock.callsFake(repo => Promise.resolve(repo))
+        mock.callsFake(repo =>
+            Promise.resolve({ name: repo, isFork: false, isArchived: false })
+        )
         const name = await resolvePackageRepo(
             '{"repository": {"url": "foo"}}',
             api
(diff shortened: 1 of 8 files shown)


extensions/typescript/src/package.test.ts
```diff
@@ -7,7 +7,9 @@ describe('resolvePackageRepo', () => {
     it('resolves string repo', async () => {
         const api = new API()
         const mock = sinon.stub(api, 'resolveRepo')
-        mock.callsFake(repo => Promise.resolve(repo))
+        mock.callsFake(repo =>
+            Promise.resolve({ name: repo, isFork: false, isArchived: false })
```
creachadair, line 11: Do we also need to cover the case where `isFork` is not defined on the object?

shared/search/providers.ts
```diff
@@ -570,3 +631,26 @@ export async function raceWithDelayOffset<T>(
...
+ * @param includeFork Whether or not the include forked repositories regardless of settings.
+ * @param includeArchived Whether or not the include archived repositories regardless of settings.
+ */
+function addRepositoryKindTerms(
+    queryTerms: string[],
+    includeFork: boolean,
+    includeArchived: boolean
+): string[] {
+    if (includeFork || getConfig('basicCodeIntel.includeForks', false)) {
+        queryTerms.push('fork:yes')
```
creachadair, line 648: The API is a little odd here in that we are modifying the input argument but also returning it. The usage site makes it look like we're getting a copy, but we're not really—the original has been tweaked.

I'd just let it modify in-place and document it as such, probably.

shared/util/api.ts
```diff
@@ -42,28 +42,61 @@ export interface LineMatch {
...
-            repository: { name: string }
+            repository: RepoMeta
         }
 
         const data = await queryGraphQL<Response>(query, { cloneURL })
-        return data.repository.name
+        return data.repository
+    }
+
+    /** Determines via introspection if the GraphQL API has iSFork field on the Repository type. */
```
creachadair, line 81: Is this needed so that we can safely update the client before we update the GraphQL server implementation?
If so, then I assume that means we can get rid of this eventually—in which case could you please add a TODO to that effect somewhere nearby?

creachadair (commented):
Just a handful of small questions.

extensions/typescript/src/package.test.ts
efritz: It's always defined (the API.resolveRepo will return false for isFork/isArchived if introspection shows the fields aren't defined on older instances).

shared/util/api.ts
efritz: I'm not sure what our contract for old instances are. We should decide that as well, but I'll add a comment here just in case we can eventually remove it.

shared/search/providers.ts
efritz: Yeah that was dirty. Updated.

creachadair (approved):

shared/util/api.ts
```diff
@@ -42,28 +42,65 @@ export interface LineMatch {
...
+
+    /**
+     * Determines via introspection if the GraphQL API has iSFork field on the Repository type.
+     *
+     * TODO(efritz) - Remove this when we no longer need to support pre-3.15 instances.
+     */
+    private async hasForkField(): Promise<boolean> {
+        const introspectionQuery = gql`
+            query RepositoryIntrospection() {
+                __type(name: "Repository") {
```
felixfbecker, line 89: This is really cool

efritz: ikr

PR #366, opened 2020-04-08
resolveRepo: use name, not clone URL

lguychard:
With https://github.com/sourcegraph/code-intel-extensions/pull/363, basic code intel now calls `resolveRepo()` when creating providers. `resolveRepo()` is called with the repo name, but passes it as the cloneURL in graphQL queries: https://sourcegraph.com/search?q=repo:%5Egithub%5C.com/sourcegraph/code-intel-extensions%24+%5CWresolveRepo%28&patternType=regexp

This works for repo names of the form `github.com/sourcegraph/jsonrpc2` (although I think that's by accident, not by design):

image

Unfortunately it fails in many cases (eg. when a `repositoryPathPattern` excluding the hostname from the repo name is set, or for single git repositories), causing code intel to break:

image

image

Changing the graphQL query to query by repo name fixes things:

image

### DIFF
--- a/shared/util/api.ts
+++ b/shared/util/api.ts
@@ -56,14 +56,14 @@ export class API {
      *
      * @param cloneURL The repository's clone URL.
      */
-    public async resolveRepo(cloneURL: string): Promise<RepoMeta> {
+    public async resolveRepo(name: string): Promise<RepoMeta> {
         const metaFields = (await this.hasForkField())
             ? 'isFork\nisArchived'
             : ''
 
         const query = gql`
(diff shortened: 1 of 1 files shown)


shared/util/api.ts
```diff
@@ -56,14 +56,14 @@ export class API {
      *
      * @param cloneURL The repository's clone URL.
```
lguychard, line 57:
```suggestion
     * @param name The repository's clone name.
```

efritz (approved):

PR #370, opened 2020-04-09
Set default isFork and isArchived values on older instances

efritz:
Set defaults so that RepoMeta does not have undefined values for boolean fields.


creachadair:
LGTM

PR #374, opened 2020-04-29
LSIF Indexing Campaign

garobrik:
redux

### DIFF
--- a/.github/workflows/lsif.yml
+++ b/.github/workflows/lsif.yml
@@ -1,0 +1,15 @@
+name: LSIF
+on:
+  - push
+jobs:
+  lsif-ts:
+    runs-on: ubuntu-latest
+    container: sourcegraph/lsif-node
+    steps:
+      - uses: actions/checkout@v1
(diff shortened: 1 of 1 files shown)


efritz (changes requested):
Still does not index multi-project

gbrik:
as it turns out, this was the only repo which had this issue. going to fix manually in the morning

macraig:
@gbrik @efritz can this be closed?

efritz:
This doesn't currently work because lsif-tsc has an issue with multi-projects. We can close it, but leaving it opens serves as a good reminder to get this to work once we have the bandwidth.

Strum355:
@efritz am i misremembering that lsif-tsc now works for multi-projects?

efritz:
@Strum355 Could you try running the master version of the indexer and see if it works? If so we can update the args here.

Strum355:
`lsif-tsc -p .` discovered no files. `lsif-tsc **/*.ts` did work

PR #383, opened 2020-05-07
Add TCL and Cobol extensions

efritz:


shared/language-specs/languages.ts
```diff
@@ -28,6 +28,14 @@ const clojureSpec: LanguageSpec = {
     commentStyles: [lispStyleComment],
 }
 
+const cobolSpec: LanguageSpec = {
+    languageID: 'cobol',
+    stylized: 'Cobol',
+    fileExts: ['cbl', 'cob', 'cpy', 'dds', 'ss', 'wks', 'pco'],
+    identCharPattern: createIdentCharPattern('-'),
```
gbrik, line 35: question:
in general how does this handle which chars can be first in an identifier and which can only appear after the first char?

shared/language-specs/languages.ts
```diff
@@ -28,6 +28,14 @@ const clojureSpec: LanguageSpec = {
...
+    commentStyles: [{ lineRegex: /\*/ }],
```
gbrik, line 36: question:

The doc for `lineRegex` says:

```
Captures the content of a line comment. Also prevents jump-to-definition
(except when the token appears to refer to code). Python example:
/#\s?(.*)/
```

This includes a capture group for the comment content. But this regex, and also `shellStyleComment` and `cStyleComment` don't include a capture group. Is the doc example wrong, or are both valid?

garobrik (approved):

shared/language-specs/languages.ts
efritz: Right now this is only used to scan left/right from the current hover position eating characters while it can on each side. We don't currently support a way to make special first characters (java idents can't start with a number, for example, but we'd match it since numbers are in the regex).

shared/language-specs/languages.ts
efritz: Good question. I'll look into this a bit before merging to make this more clear.

shared/language-specs/languages.ts
gbrik: so basically this would consider an integer literal an identifier, but that doesn't actually affect usability?

efritz: Yes, on defs it won't match any output of ctags, and without a def you can't find references either.

If you'd like you can open an issue to improve identifier selection. There's probably a lot of little heuristic tweaks we can do to improve this.

PR #385, opened 2020-05-14
Respect version contexts in search providers

efritz:
Pass versionContext into search functions if it is available to the workspace root.

### DIFF
--- a/shared/util/api.ts
+++ b/shared/util/api.ts
@@ -347,11 +347,17 @@ export class API {
         searchQuery: string,
         fileLocal = true
     ): Promise<SearchResult[]> {
+        const context = sourcegraph.workspace.versionContext
+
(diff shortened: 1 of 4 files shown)


garobrik (approved):

shared/util/api.ts
```diff
@@ -406,7 +412,12 @@ export class API {
             }
         }
 
-        const data = await queryGraphQL<Response>(query, { query: searchQuery })
+        let vars: { [name: string]: unknown } = { query: searchQuery }
+        if (context) {
+            vars = { context, ...vars }
+        }
+
+        const data = await queryGraphQL<Response>(query, vars)
```
felixfbecker, line 420: This should work since if `context` is `undefined`, it'll be omitted from the JSON object
```suggestion
        const data = await queryGraphQL<Response>(query, { query: searchQuery, context })
```

felixfbecker (approved):

PR #391, opened 2020-05-19
Update conditions to check for empty arrays in top-level providers

efritz:
We were emitting `codeintel.searchDefinitions` events each time the provider was active and didn't return a precise result.

This PR checks explicitly for empty arrays and guards the emission of events and results when there's nothing interesting to emit.

### DIFF
--- a/shared/util/helpers.ts
+++ b/shared/util/helpers.ts
@@ -8,6 +8,18 @@ export function isDefined<T>(value: T | undefined): value is T {
 }
 
 /**
+ * Returns true fi the value is defined and, if an array, contains at least
+ * one element.
+ *
+ * @param value The value to test.
+ */
+export function nonEmpty<T>(
+    value: T | T[] | null | undefined
+): value is T | T[] {
+    return !!value && !(Array.isArray(value) && value.length === 0)
+}
+
+/**
  * Ensure that the given value is an array.
  *
  * @param value The list of values, a single value, or null.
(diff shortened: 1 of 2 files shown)


shared/util/helpers.ts
```diff
@@ -7,6 +7,18 @@ export function isDefined<T>(value: T | undefined): value is T {
     return value !== undefined
 }
 
+/**
+ * Returns true fi the value is defined and, if an array, contains at least
```
gbrik, line 11: typo: fi -> if

shared/util/helpers.ts
```diff
@@ -7,6 +7,18 @@ export function isDefined<T>(value: T | undefined): value is T {
...
+ * one element.
+ *
+ * @param value The value to test.
+ */
+export function nonEmpty<T>(
+    value: T | T[] | null | undefined
+): value is T | T[] {
+    return !!value && !(Array.isArray(value) && value.length === 0)
```
gbrik, line 19: question:
wut sorcery is this? what is the `value is T | T[]` return type, and why `!!value`?

garobrik (approved):

shared/util/helpers.ts
efritz: `value is X` is a type assertion so that at the call site it will refine the type of the parameter inside the conditional.

`!!x` is equivalent to `Boolean(x)` - it negates the truthiness twice to make a boolean. Without it the expression returns bool or undefined, not bool (due to short circuiting of &&).

PR #414, opened 2020-07-01
Batch LSIF queries via window resolver

efritz:
This PR is a backwards-compatible change with how LSIF queries work. Previously, we would perform two LSIF queries on each hover action: one for definition and one for hover text. When a user scrolls over the page this can make a lot of network requests. A way to batch queries was introduced in https://github.com/sourcegraph/sourcegraph/pull/11815. This adds a GraphQL resolver that will return all ranges along with their local definitions, references, and hover text within a window of lines for a document.

We now query a bit differently: we maintain a set of "windows", which is a start line, end line, and an array of ranges contained within that window. When looking data for a position, we first check to see if we have a window containing that position. If we don't, we make a batch query for that position +50/-50 lines and store it in memory. We then look inside that window for the range that contains the position and returns the associated data.

The data returned by batch queries is not complete: it does not look outside of the bundle to find cross-root or cross-repository results. It is still necessary to make position-specific queries to resolve the data that is missing from the batch queries.

This reduces the number of network requests in the common case: instead of performing two requests per hover, we only perform one request to resolve all local hover text and definitions within a visible region. Until the user scrolls 50 lines away and hovers, there will be no other network requests (except for xroot/xrepo definitions and reference requests, which is not meant to be an instantaneous operation).

This implements the client-side of https://github.com/sourcegraph/sourcegraph/issues/10869.


shared/lsif/api.ts
```diff
@@ -0,0 +1,91 @@
...
+                lsif {
+                    ${inner}
+                }
+            }
+        }
+    }
+`
+
+export const resourceFragment = gql`
+    resource {
```
felixfbecker, line 34: You could use the GraphQL fragment syntax here, which makes this string formattable by Prettier and even validatable & autocompletable (although I think that tooling is not setup  in this repo). Still makes it explicit on what type these fields apply to :)

https://graphql.org/learn/queries/#fragments

shared/lsif/api.ts
```diff
@@ -0,0 +1,91 @@
...
+
+/** The response envelope for all LSIF queries. */
+export interface GenericLSIFResponse<R> {
+    repository: {
+        commit: {
+            blob: {
+                lsif: R
+            }
+        }
+    }
```
felixfbecker, line 13: Can any of these be `null`?

shared/lsif/definition.ts
```diff
@@ -0,0 +1,67 @@
...
+                line: position.line,
+                character: position.character,
+            },
+            queryGraphQL
+        )
+    )
+}
+
+/**
+ * Convert a GraphQL definition response into a list of Sourcegraph lcoations.
```
felixfbecker, line 55:
```suggestion
 * Convert a GraphQL definition response into a list of Sourcegraph locations.
```

shared/lsif/ranges.ts
```diff
@@ -0,0 +1,350 @@
...
+export async function makeRangeWindowFactory(
+    queryGraphQL: QueryGraphQLFn<any> = sgQueryGraphQL
+): Promise<RangeWindowFactoryFn> {
+    if (!(await hasRangesQuery(queryGraphQL))) {
+        // No-op if the instance doesn't support bulk loading
+        return () => Promise.resolve(null)
+    }
+
+    const cache = new Map<
+        sourcegraph.TextDocument,
```
felixfbecker, line 75: You may wanna store by the URI string just in case a text document is the same URI but not referentially equal (they _should_ generally be, unless a text document gets removed, and then added later again)

felixfbecker (commented):
Didn't go deep on the logic since this is a lot of code, but nothing jumped out to me.

shared/lsif/api.ts
efritz: I'm making them nullable to be safe (the previous assumption was that we're _on_ the blob, so it shouldn't be).

shared/lsif/definition.ts
```diff
@@ -0,0 +1,67 @@
...
+    lsifRequest,
+} from './api'
+
+export interface DefinitionResponse {
+    definitions: {
+        nodes: LocationConnectionNode[]
+    }
+}
+
+const definitionsQuery = gql`
```
twop, line 19: I don't know if there is a way to typecheck this, but it seems it would be really hard to debug if there is a typo or api change, what do you think?

twop (approved):
I skimmed through the diff and it seems ok. It is hard for me to assess such a big change without context. I you want a specific piece of code that you would to be reviewed please let me know.

shared/lsif/api.ts
efritz: Updated, added tooling.

shared/lsif/definition.ts
efritz: Added CI steps to ensure that the schema is up to date and the queries are validated against the schema.

shared/lsif/references.ts
```diff
@@ -0,0 +1,154 @@
...
+                        }
+                    }
+                }
+            }
+        }
+    }
+`
+
+/** Retrieve references for the current hover position. */
+// eslint-disable-next-line @typescript-eslint/require-await
```
felixfbecker, line 74: Btw, this issue is fixed if you upgrade to the latest `@sourcegraph/eslint-config` version :)

https://github.com/typescript-eslint/typescript-eslint/issues/1691

felixfbecker (approved):
Cool!

shared/lsif/references.ts
efritz: Will deal with a bunch of renovate stuff next.

dev/scripts/update-graphql-schema.ts
```diff
@@ -0,0 +1,17 @@
+import * as fs from 'mz/fs'
+import fetch from 'node-fetch'
+
+const SCHEMA_URL =
+    'https://raw.githubusercontent.com/sourcegraph/sourcegraph/master/cmd/frontend/graphqlbackend/schema.graphql'
+
+async function main(): Promise<void> {
+    await fs.writeFile(
+        './schema/schema.graphql',
+        await (await fetch(SCHEMA_URL)).text()
```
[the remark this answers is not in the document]
efritz: curl ain't ts

.buildkite/check-graphql-schema.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -ex
+cd "$(dirname "${BASH_SOURCE[0]}")"
+
+./yarn-run.sh "update-graphql-schema"
```
felixfbecker, line 6: Doesn't this make the build non-deterministic/non-reproducible if the schema changes in the main repo?

efritz: Yeah, in which case we should ensure the extension still works because we need to support all instance behaviors for all time forever.

felixfbecker: It still doesn't seem like a good idea to sacrifice build reproducability for that. It would mean _every_ PR is suddenly failing, even a PR to e.g. revert a commit.
If you just want to be notified of any changes to the schema, maybe set up a Sourcegraph saved search with notifications for the schema file in our main repo?

PR #421, opened 2020-07-14
Do not block on range requests

efritz:
Range requests can take a while to resolve if there is a lot of data in a window. This is especially problematic in symbol-dense code such as Typescript (https://sourcegraph.com/github.com/sourcegraph/sourcegraph@138e2f89a0136fab21d5f76a981c38e686f281d9/-/blob/web/src/search/input/MonacoQueryInput.tsx#L193:13).

This changes the providers to request a range from the API, but only take a short 25ms delay until requesting the definition or hover for the position explicitly. The range request will continue to resolve in the background so once it's in memory all other requests will no longer need to make explicit requests for local data.

### DIFF
--- a/shared/lsif/ranges.ts
+++ b/shared/lsif/ranges.ts
@@ -180,8 +180,8 @@ export function findOverlappingCodeIntelligenceRange(
     position: sourcegraph.Position,
     ranges: CodeIntelligenceRange[]
 ): CodeIntelligenceRange | null {
-    return (
-        ranges.find(
+    const overlapping =
+        ranges.filter(
             ({
                 range: {
                     start: { line: startLine, character: startCharacter },
@@ -193,7 +193,21 @@ export function findOverlappingCodeIntelligenceRange(
                 // right side check
                 (position.line < endLine || (position.line === endLine && position.character < endCharacter))
         ) || null
-    )
+
+    if (overlapping.length === 0) {
+        return null
+    }
+
+    overlapping.sort((a, b) => {
+        const cmp = b.range.start.line - a.range.start.line
+        if (cmp === 0) {
+            return b.range.start.character - a.range.start.character
+        }
+
+        return cmp
+    })
+
+    return overlapping[0]
 }
 
 const introspectionQuery = gql`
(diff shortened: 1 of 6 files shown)


shared/util/promise.ts
```diff
@@ -0,0 +1,45 @@
...
+ * @param makeFallback A factory that creates a fallback promise.
+ * @param timeout The timeout in ms before the fallback is invoked.
+ * @param filter An optional filter function to determine if a set of results
+ *        should be returned immediately.
+ */
+export async function raceWithDelayOffset<T>(
+    primary: Promise<T>,
+    makeFallback: () => Promise<T>,
+    timeout: number,
+    filter: (v: T) => boolean = v => v !== null
```
eseliger, line 22: Wasn't the original behavior `!== undefined`?

shared/lsif/ranges.ts
```diff
@@ -193,7 +193,21 @@ export function findOverlappingCodeIntelligenceRange(
...
+    }
+
+    overlapping.sort((a, b) => {
+        const cmp = b.range.start.line - a.range.start.line
+        if (cmp === 0) {
+            return b.range.start.character - a.range.start.character
+        }
+
+        return cmp
+    })
```
felixfbecker, line 208: You may be able to make use of our extension APIs - if you convert these to `sourcegraph.Range`/`sourcegraph.Position` you have methods like `compareTo()`, `isBefore()`, `contains()` etc

eseliger (approved):

shared/util/promise.ts
efritz: Yes but that's only for the race against a delay promise, which returns undefined. This was only used in unindexed/indexed search fallback, and I'm sure that not returning null is okay in that circumstance as well.

shared/util/promise.ts
```diff
@@ -0,0 +1,45 @@
...
+): Promise<T> {
+    const primaryResults = await Promise.race([primary, delay(timeout)])
+    if (primaryResults !== undefined && filter(primaryResults)) {
```
felixfbecker, line 25: I find it a bit surprising that `undefined` is implicitly handled but `null` is only handled by the `filter` function

PR #422, opened 2020-07-14
Fix "Invalid array length" bug with spread of set values

efritz:
This should fix the issue reported in https://sourcegraph.atlassian.net/jira/servicedesk/projects/SG/queues/custom/1/SG-267.

### DIFF
--- a/shared/search/providers.ts
+++ b/shared/search/providers.ts
@@ -527,9 +527,9 @@ function sortByProximity(locations: sourcegraph.Location[], currentURI: URL): so
 function jaccardIndex<T>(a: Set<T>, b: Set<T>): number {
     return (
         // Get the size of the intersection
-        new Set([...[...a]].filter(value => b.has(value))).size /
+        new Set([...a].filter(value => b.has(value))).size /
         // Get the size of the union
-        new Set([...[...a], ...[...b]]).size
+        new Set([...a, ...b]).size
     )
 }
 
--- a/shared/tsconfig.json
+++ b/shared/tsconfig.json
@@ -9,6 +9,7 @@
         "tag": "gql",
       },
     ],
+    "downlevelIteration": true,
   },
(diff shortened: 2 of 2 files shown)


shared/search/providers.ts
```diff
@@ -527,9 +527,9 @@ function sortByProximity(locations: sourcegraph.Location[], currentURI: URL): so
 function jaccardIndex<T>(a: Set<T>, b: Set<T>): number {
     return (
         // Get the size of the intersection
-        new Set([...[...a]].filter(value => b.has(value))).size /
```
twop, line 530: Now I'm really curios why it was the way it was and what made it go  away? Tanspiling?

efritz: https://github.com/sourcegraph/code-intel-extensions/pull/384/files#diff-d7d6c42b22b4cc3c9e90d696d64e26eeR530 🤷

felixfbecker: This is really interesting because while the code before was already double-copying the arrays (not good), the autofixed double spread looks equivalent and not like it shouldn't work or cause the RangeError 🤔

shared/tsconfig.json
```diff
@@ -9,6 +9,7 @@
         "tag": "gql",
       },
     ],
+    "downlevelIteration": true,
```
felixfbecker, line 12: Why is this needed? We only target modern browsers so generally shouldn't do this as it could bloat bundles. Is this significant to fix the bug?

efritz: This is the only thing that fixes the bug. The double spread is just a drive-by.

shared/search/providers.ts
efritz: This didn't fix the error, I just cleaned this up since I noticed it. I have no clue why it was doing a double copy in the first place.

PR #423, opened 2020-07-15
hover source code intel tooltips

garobrik:
TODO: update package.json to point to new sourcegraph extension API when it's published.

product: see the new hoverAlerts.ts file if all you're interested in is the text. you can see what the alert will actually look like on https://github.com/sourcegraph/sourcegraph/pull/12123. I want to make the alerts for search and lsp-based results not dismissible when a robust LSIF indexer is available for the languages, but open to pushback if you think that's too obnoxious.


shared/providers.ts
```diff
@@ -298,7 +305,11 @@ export function createHoverProvider(
             for await (const lsifResult of lsifProvider(textDocument, position)) {
                 if (lsifResult) {
                     await emitter.emitOnce('lsifHover')
-                    yield lsifResult
+                    if (hasPreciseResult) {
+                        yield lsifResult
+                    } else {
+                        yield { ...lsifResult, alerts: HoverAlerts.LSIF }
```
efritz, line 311: Why are we not tagging the first lsif result?

shared/providers.ts
```diff
@@ -309,10 +320,19 @@ export function createHoverProvider(
 
             if (lspProvider) {
                 // Delegate to LSP if it's available.
-                for await (const lspResult of lspProvider(textDocument, position)) {
+                let hasLSPResult = false
```
efritz, line 323: You'll need to run prettier in order for ci to pass.

shared/hoverAlerts.ts
```diff
@@ -0,0 +1,84 @@
+import * as sourcegraph from 'sourcegraph'
+
+export const HoverAlerts = {
+    LSIF: <sourcegraph.Badged<sourcegraph.HoverAlert>[]>[
```
felixfbecker, line 4: This casts the expression to the type, which means it silences any errors of missing properties. If you want to keep the `HoverAlerts` namespace, I would instead suggest to just do
```ts
export const lsif: sourcegraph.Badged<sourcegraph.HoverAlert>[] = [
  // ...
]
```
and import with
```ts
import * as HoverAlerts from './hoverAlerts'
```
(the difference is casting vs a type annotation)

There's an ESLint rule that shows this as a warning if you have some ESLint integration installed for your editor.

shared/hoverAlerts.ts
```diff
@@ -0,0 +1,84 @@
...
+        {
+            summary: {
+                kind: sourcegraph.MarkupKind.Markdown,
+                value:
+                    'Semantic result. [Learn more.](https://docs.sourcegraph.com/user/code_intelligence/lsif)',
+            },
+            badge: {
+                kind: 'info',
+                hoverMessage:
+                    "This hover text comes from a pre-computed semantic index of this project's source. Click to learn how to add this capability to all of your projects!",
```
christinaforney, line 14: 'hover text' may not be clear or accurate. Alternative ideas: 
- "This code intelligence data"
- "This hover data"
- "This reference data"

We have some style guide info on wording of links, though this is a bit difference since it's in a hover: https://about.sourcegraph.com/handbook/communication/style_guide#links

shared/hoverAlerts.ts
```diff
@@ -0,0 +1,84 @@
+import * as sourcegraph from 'sourcegraph'
+
+export const HoverAlerts = {
+    LSIF: <sourcegraph.Badged<sourcegraph.HoverAlert>[]>[
+        {
+            summary: {
+                kind: sourcegraph.MarkupKind.Markdown,
+                value:
+                    'Semantic result. [Learn more.](https://docs.sourcegraph.com/user/code_intelligence/lsif)',
```
christinaforney, line 9: In context on the hover I think these make sense, but reading them here I'm wondering if we could be more clear what this means. Not everyone will understand what a language server is, what semantic means in this context, etc. I know some of this will be clarified in the hover info, but am curious to briefly discuss.

christinaforney (commented):
It would be great if you could get some time with @LaureenH to show her how this text shows up in the product and get her feedback!

shared/hoverAlerts.ts
LaureenH: maybe just "text"

LaureenH: Also, I love that you ended with an exclamation mark. This is *exactly* the kind of enthusiasm I'm trying to get you to express in the release blog content. =)

christinaforney: @LaureenH - the reason I made alternate suggestions is that "text" feels inaccurate to me. This is because it'ts not just copy. It's information about your code including the function definition, parameters and types, as well as any available documentation in the code. Here's what it looks like: 
image

shared/providers.ts
gbrik: we're tagging *only* the first lsif result. is the var name confusing?

efritz: Ah I see. Why not tag all results? I'm not sure how this is read by the webapp, but I'm assuming there's an implicit piece of knowledge about how it's displayed.

shared/hoverAlerts.ts
gbrik: ahh ok, will change.

for fun: before i added the casts, i was trying to create an object with every key of type Badged<HoverAlert>[] so I didn't have to write it every time. but Record<string, Badged<HoverAlert>[]> didn't work. do you know if there's a way to do this pattern in TS? basically a Map<K, V> with compile time enforced finite Ks?

shared/hoverAlerts.ts
gbrik: other ideas:
- 'content'/'hover content'
- 'symbol information'

I favour 'text' because I like concision, or 'symbol information' for maximum specificity, thoughts?

shared/hoverAlerts.ts
gbrik: There's a thread here where we were trying to brainstorm how to get this across, not much luck: https://sourcegraph.slack.com/archives/CHXHX7XAS/p1589989321162100

The text before 'Learn more.' would need to be max ~4 words for the alert to fit on one line, which I would really like.

shared/hoverAlerts.ts
gbrik: also, my style preference is to have constants start with upper case. my impression from our TS source is that types are uppercase, everything else is lowercase. is that a distinction we care about?

shared/providers.ts
gbrik: The webapp does not ATM dedup alerts, so if every result has an alert on it it'll show multiple alerts. I agree there's probably a better way to do this, but not sure what it is. This is another artifact of designing through the extension API :(

efritz (approved):

shared/hoverAlerts.ts
gbrik: I went with 'hover data', symbol information is too specific, and the hovers in general could contain content other than text.

shared/hoverAlerts.ts
christinaforney: Agree, and I remember that thread. This seems like the best we have for now!

christinaforney (approved):

PR #432, opened 2020-07-27
Combine LSIF definition and hover providers

### LINKED ISSUE #12133: Query definition and hover provider so we can correctly badge results
In the code intel extensions, the definition may be imprecise but a hover may be precise (this occurs with go stdlib values where definitions aren't resolvable via LSIF the hovers are). We need to query the definition along with the hover in order to send back a properly badged value.

efritz:
Closes https://github.com/sourcegraph/sourcegraph/issues/12133.

This creates a new provider type `CombinedProviders` which will query both definitions and hovers in the same request rather than individually. This allows us in the hover providers that falls back to search-based results to know whether or not we have precise or imprecise definitions. This allows us to tag the scenarios where we have a precise hover and imprecise definition, and have an imprecise hover and a precise definition. The former happens on references to the go stdlib, for example, which is very jarring as there is no UI hint that you'll be taken to an incorrect definition.


shared/lsif/providers.ts
```diff
@@ -39,50 +37,94 @@ export function createProviders(logger: Logger): Providers {
...
-        definition: asyncGeneratorFromPromise(definition(queryGraphQL, getRangeFromWindow)),
+        definitionAndHover: cachedDefinitionAndHover(queryGraphQL, getRangeFromWindow),
         references: references(queryGraphQL, getRangeFromWindow),
-        hover: asyncGeneratorFromPromise(hover(queryGraphQL, getRangeFromWindow)),
         documentHighlights: asyncGeneratorFromPromise(documentHighlights(queryGraphQL, getRangeFromWindow)),
     }
 }
 
 /** The time to delay between range queries and an explicit definition/reference/hover request. */
 const RANGE_RESOLUTION_DELAY = 25
```
aidaeology, line 49: Nit: I'm assuming this is milliseconds, but it would be better to have explicit descriptors for the unit either in the variable name i.e. RANGE_RESOLUTION_DELAY_MS or RANGE_RESOLUTION_DELAY_SECS or in the comment section: "/** time to delay between range queries in milliseconds.."

shared/lsif/definition-hover.test.ts
```diff
@@ -0,0 +1,95 @@
...
+            hover: null,
+        })
+    })
+
+    it('should deal with empty payload', async () => {
+        const queryGraphQLFn = sinon.spy<QueryGraphQLFn<GenericLSIFResponse<DefinitionAndHoverResponse | null>>>(() =>
+            makeEnvelope()
+        )
+
+        assert.deepEqual(await definitionAndHoverForPosition(document, position, queryGraphQLFn), null)
```
aidaeology, line 93: Question here on the behavior: does this mean that when definition & hover text are not found (empty) that we display to the user a hover UI that is blank or is there someplace else in the stack where we populate a "not found" message to inform the user vs blank response?

efritz: If there's a definition and no hover, we show buttons but no text (I think - it's possible we have canned text, or we may not display the buttons in this case either).

If there's a hover but no def, we just don't show action buttons (that I know for sure).

aidaeology: Ok just wanted to make sure it was handled somewhere.

shared/lsif/providers.ts
efritz: Taken care of in https://github.com/sourcegraph/code-intel-extensions/pull/433.

shared/providers.ts
```diff
@@ -301,21 +321,22 @@ export function createHoverProvider(
...
-            let hasPreciseResult = false
-            for await (const lsifResult of lsifProvider(textDocument, position)) {
-                if (lsifResult) {
-                    await emitter.emitOnce('lsifHover')
-                    if (hasPreciseResult) {
-                        yield lsifResult
-                    } else {
-                        yield { ...lsifResult, alerts: HoverAlerts.lsif }
+            const lsifWrapper = await lsifProvider(textDocument, position)
+            if (lsifWrapper) {
```
aidaeology, line 325: My understanding is that LSIF functionality replaces LSP - is it a common use case where we have both LSP and LSIF providers for a language on one instance? If we don't expect that to be true then this logic could just be an either/or; where first check for optional LSP provider results in true, skip check for LSIF. If no LSP then try LSIF.

aidaeology, line 325: My understanding is that LSIF functionality replaces LSP - is it a common use case where we have both LSP and LSIF providers for a language on one instance? If we don't expect that to be true then this logic could just be an either/or; where first check for optional LSP provider results in true, skip check for LSIF. If no LSP then try LSIF.

efritz: It is possible to have both for Go or TypeScript, which is unfortunate due to maintenance complexity. Once we're able to ensure that most of our large private instances have a path forward with LSIF we can sunset the LSP support in the extension completely.

aidaeology: Ok thanks for the context.

aidaeology (approved):

Issue #475, opened 2020-09-03
Partial semantic result tooltip language


efritz:
This reads strangely if there is no goto def available at all. This tooltip should not be shown unless there is _a_ non-precise definition.

PR #476, opened 2020-09-04
Simplify provider caches

efritz:
This is to clean up a bit before https://github.com/sourcegraph/code-intel-extensions/issues/475, which requires that we cache closer to the original providers in order to effectively determine the information we need without recomputing multiple times per request.

This moves the search cache closer to the search provider for definitions, rather than memoizing the combined LSIF/LSP/Search providers.

### DIFF
--- a/shared/util/ix.ts
+++ b/shared/util/ix.ts
@@ -124,7 +124,7 @@ export const PROMISE_CACHE_CAPACITY = 5
  *
  * @param func The promise function.
  */
-export function cacheProviderPromise<P extends unknown[], R>(
+export function cachePromiseProvider<P extends unknown[], R>(
     func: (...args: P) => Promise<R>,
     cacheCapacity: number = PROMISE_CACHE_CAPACITY
 ): (...args: P) => Promise<R> {
--- a/shared/util/ix.ts
+++ b/shared/util/ix.ts
@@ -119,9 +119,10 @@ export function asyncGeneratorFromPromise<P extends unknown[], R>(
 export const PROMISE_CACHE_CAPACITY = 5
 
 /**
- * TODO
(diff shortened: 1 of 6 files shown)


shared/util/ix.ts
```diff
@@ -113,3 +114,56 @@ export function asyncGeneratorFromPromise<P extends unknown[], R>(
         yield await func(...args)
     }
 }
+
+/** The maximum number of promise results to cache. */
+export const PROMISE_CACHE_CAPACITY = 5
```
gbrik, line 119: question:
any context for why this is 5?

shared/util/ix.ts
```diff
@@ -113,3 +114,56 @@ export function asyncGeneratorFromPromise<P extends unknown[], R>(
...
+
+/**
+ * Memoizes a function that returns a promise. Internally, this maintains a simple
+ * bounded LRU cache.
+ *
+ * @param func The promise function.
+ */
+export function cachePromiseProvider<P extends unknown[], R>(
+    func: (...args: P) => Promise<R>,
```
gbrik, line 128: question:
can this be more typesafe? we cast `P` to `[sourcegraph.TextDocument, sourcegraph.Position]` later but my typescript fu is not good enough to know if it's easy to propogate that to this signature

garobrik (approved):
LGTM

shared/util/ix.ts
efritz: It's mainly to cache back-to-back requests (we want a search definition like 2-3 times per hover and we don't want to recalculate two of those three). Five is just a bit of headroom.

shared/util/ix.ts
efritz: Maybe but I can't type it correctly so I'm not getting yelled at by some langserver somewhere.

PR #477, opened 2020-09-04
Fix partial results tooltip condition

### LINKED ISSUE #475: Partial semantic result tooltip language

efritz:
This PR stops the "definition may be imprecise" tooltip when there is no definition _to_ be imprecise. Closes https://github.com/sourcegraph/code-intel-extensions/issues/475.

### DIFF
--- a/shared/lsif/providers.ts
+++ b/shared/lsif/providers.ts
@@ -1,7 +1,7 @@
 import * as sourcegraph from 'sourcegraph'
 import { noopProviders, CombinedProviders, DefinitionAndHover } from '../providers'
 import { queryGraphQL as sgQueryGraphQL, QueryGraphQLFn } from '../util/graphql'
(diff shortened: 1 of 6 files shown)


aidaeology (approved):
LGTM

Strum355 (approved):
https://github.com/sourcegraph/code-intel-extensions/pull/477/files?diff=split&w=1#diff-51a05538570b924ce55ee8d638480344R386 curious why the change (or why searchHoverProvider wasnt used before), but otherwise :shipit:

efritz:
@Strum355 I see the confusion. We were using the search hover provider (previously called `searchProvider`) to get the hover text. We added a search definition provider so that we only show the text "Hover text is precise but definition isn't" when there's actually an imprecise definition.

PR #480, opened 2020-09-06
Use standard URL in package.json repository object

sqs:
The `github:OWNER/REPO` syntax might be handled specially by some tools, but it's non-standard. It also breaks the `Repository` link on the extension's page at (eg) https://sourcegraph.com/extensions/sourcegraph/go.


efritz (approved):
Are the same links in the extensions/**/package.json files ok?

PR #487, opened 2020-09-10
Add setting to disable range queries

efritz:
Adding `"codeIntel.disableRangeQueries": true` to user settings now disables range queries, which can be very expensive in older instances. I'm adding this to direct pre-3.20 customers with certain poorly performing range queries as a workaround until they can upgrade.


garobrik (approved):
the dup'd code makes me :cry:

efritz:
It won't be there once we remove LSP support.

PR #488, opened 2020-09-10
Do not invoke a search unnecessarily

efritz:
This should reduce the number of search queries for repositories with LSIF data.


garobrik (changes requested):
can i has test?

efritz:
@gbrik there is your STINKY test

garobrik (approved):
smells delicious :nose:

PR #506, opened 2020-10-09
Remove samples directory

efritz:


Strum355 (approved):

felixfbecker:
@marekweb we would need to update the commit hash in the main repo after this. Is that safe or does it have any implications on our current Firefox review?

marekweb:
It's safe to update the main repo. The review won't be affected because it doesn't include newer commits to the main repo.

PR #508, opened 2020-10-22
Update doc links to the new doc schema

LawnGnome:
I noticed this when I clicked on the info link in this hover:

image

It pointed to https://docs.sourcegraph.com/user/code_intelligence/explanations/precise_code_intelligence, which no longer exists.

This is exacerbated by some missing redirects in our doc site, which I'm going to open a PR for in https://github.com/sourcegraph/sourcegraph, but we should obviously fix this in _all the places_. :broom:


efritz (approved):
Thank you!

PR #528, opened 2020-11-10
Fix handling of repositories with spaces in repo or path

### LINKED ISSUE #526: Basic code intel search query that gets generated does not %-decode repo name in repo: filter
Test on https://k8s.sgdev.org/sourcegraph.visualstudio.com/Test%20Repo/-/blob/Test%20Folder/Test%20File.cs#L7:20

The GraphQL request contains this query:

```json
query: "^ThisShouldBeUniqueEnough$ type:symbol patternType:regexp case:yes file:\.(cs|csx)$ repo:^sourcegraph.visualstudio.com/Test%20Repo$@15883869af3bc85bb5ee93880ea4014a08c1f37d"
```

Note how the `Test%20Repo` is in the `repo:` filter. It should be `Test Repo`.

This is the cause of no code intel reported in https://github.com/sourcegraph/sourcegraph/issues/15402 (there may be more, but this is definitely one of them)

The structural fix for this is https://github.com/sourcegraph/sourcegraph/issues/4278, but we need a fix for existing instances regardless.

I made a similar fix to git-blame that may be relevant as an example: https://github.com/sourcegraph/sourcegraph-git-extras/pull/155/files

### LINKED ISSUE #527: Cannot read property commit of null when repo has spaces in name
image

Reported in https://github.com/sourcegraph/sourcegraph/issues/15402

You can reproduce here: https://k8s.sgdev.org/sourcegraph.visualstudio.com/Test%20Repo/-/blob/Test%20Folder/Test%20File.cs#L7:20&tab=references

efritz:
Closes https://github.com/sourcegraph/code-intel-extensions/issues/527. Closes https://github.com/sourcegraph/code-intel-extensions/issues/526.

### DIFF
--- a/shared/search/providers.test.ts
+++ b/shared/search/providers.test.ts
@@ -19,8 +19,14 @@ const spec: LanguageSpec = {
     filterDefinitions: <T extends Result>(results: T[]) => results.filter(result => result.file !== '/f.ts'),
 }
 
-const textDocument = createStubTextDocument({
+const textDocument1 = createStubTextDocument({
     uri: 'git://sourcegraph.test/repo?rev#/foo.ts',
+    languageId: 'typescript',
+    text: undefined,
+})
+
+const textDocument2 = createStubTextDocument({
+    uri: 'git://sourcegraph.test/repo%20with%20spaces?rev#/foo.ts',
     languageId: 'typescript',
     text: undefined,
 })
@@ -135,7 +141,7 @@ describe('search providers', () => {
 
             assert.deepEqual(
                 await gatherValues(
-                    createProviders(spec, {}, api).definition({ ...textDocument, text: '\n\n\nfoobar\n' }, position)
+                    createProviders(spec, {}, api).definition({ ...textDocument1, text: '\n\n\nfoobar\n' }, position)
                 ),
                 [[new sourcegraph.Location(new URL('git://repo1?rev1#/b.ts'), range1)]]
             )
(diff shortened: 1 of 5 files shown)


Strum355 (approved):

shared/search/providers.test.ts
```diff
@@ -19,12 +19,18 @@ const spec: LanguageSpec = {
     filterDefinitions: <T extends Result>(results: T[]) => results.filter(result => result.file !== '/f.ts'),
 }
 
-const textDocument = createStubTextDocument({
+const textDocument1 = createStubTextDocument({
     uri: 'git://sourcegraph.test/repo?rev#/foo.ts',
```
felixfbecker, line 23: FYI (unrelated to the PR) paths after `#` do not have leading slashes

shared/search/providers.test.ts
```diff
@@ -563,7 +629,12 @@ describe('search providers', () => {
 })
 
 function assertQuery(searchQuery: string, expectedTerms: string[]): void {
-    const actualTerms = searchQuery.split(' ').filter(part => !!part)
+    // Split terms in a way that preserved escaped spaces
+    const actualTerms = searchQuery
+        .replace(/\\ /g, '---')
+        .split(' ')
+        .filter(part => !!part)
+        .map(term => term.replace(/---/g, '\\ '))
```
felixfbecker, line 637: FYI `split()` supports a regex and JS supports lookbehind assertions so you can do this in a non-hacky way too :) not that it's important for the test
```suggestion
    const actualTerms = searchQuery.split(/(?<!\\) /).filter(part => !!part)
```
matches only spaces not preceded by a backslash

felixfbecker (approved):
Awesome, LGTM!

felixfbecker:
Were there already tests for spaces in the file path?

shared/search/providers.test.ts
efritz: Oh I tried to do a split with regex but a capturing group and I couldn't get it not to eat the \\.

efritz:
> Were there already tests for spaces in the file path?

We don't use the path to construct queries, and we _are_ decoding the path already (we found _that_ bug earlier I guess) before sending it anywhere or comparing it with another path.

shared/search/providers.test.ts
efritz: https://github.com/sourcegraph/code-intel-extensions/pull/530

PR #530, opened 2020-11-10
Remove leading slash from fragments in tests

efritz:


Strum355:
What implementation is this change linked to? :thinking: Is it non-breaking change?

efritz:
@Strum355 Just an update to follow convention: https://github.com/sourcegraph/code-intel-extensions/pull/528#discussion_r520703315

PR #531, opened 2020-11-25
Change "basic" code intel term to "search-based"

macraig:


efritz (approved):

efritz:
We may want to change the `basicCodeIntel.*` configuration things. We can read from two variables and just have the new ones documented (so we don't change any behavior).

macraig:
Opened in [#532](https://github.com/sourcegraph/code-intel-extensions/issues/532)

PR #553, opened 2021-01-13
Do not make a search request for tokens that are in strings or comments

### LINKED ISSUE #17250: Search-based hover symbol detection is too lenient
Screen Shot 2021-01-13 at 12 39 53 PM

In the code intel extensions, we check whether or not a hovered token is within a _line_ comment, but not in block comments or strings. We need to add this detection in the extension.

efritz:
Fixes https://github.com/sourcegraph/sourcegraph/issues/17250.

We used to detect the hovered identifier's location as in a comment or not and would skip search results when it was in a comment and wasn't particularly interesting (didn't look like a projection or function call, which could be documented usage).

Now, we detect whether or not the identifier exists in block comments and strings as well. To do this we take the range +/- 50 lines around the hover area and select all text that is surrounded by that language's block comment identifiers. If the identifier exists in one of these matches, it's in a comment and we don't do a search.

Some possible unintended consequence:

- People found navigation into comment text useful (somehow)? I think this is super unlikely, and we've gotten plenty of "bogus results" feedback from weird-looking (to a human) search results.
- There are actual uses of searching in strings (e.g. field['index.thing']), where search results could be useful. We can always add rules for things like "identifiers EXACTLY in a string" or "x.{token}.y" should be matched as it's likely a unique string as we see strange behavior arise.

### DIFF
--- a/shared/search/providers.ts
+++ b/shared/search/providers.ts
@@ -104,7 +104,7 @@ export function createProviders(
             lineRegexes: commentStyles.map(style => style.lineRegex).filter(isDefined),
             identCharPattern,
         })
-        if (!tokenResult || tokenResult.isComment) {
+        if (!tokenResult || tokenResult.isString || tokenResult.isComment) {
             return undefined
         }
 
--- a/shared/search/providers.ts
+++ b/shared/search/providers.ts
@@ -102,6 +102,7 @@ export function createProviders(
             text,
             position,
             lineRegexes: commentStyles.map(style => style.lineRegex).filter(isDefined),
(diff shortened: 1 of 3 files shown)


Strum355 (approved):
:shipit:

garobrik (approved):
would also be neat-o to recognize string interpolation eventually

efritz:
@gbrik This also doesn't check for multi-line strings, so there's definitely going to be some weird code we'll encounter with a very particular string or comment contents/nesting structure that will make this just act weird.

PR #555, opened 2021-01-27
Fix $schema URL in package.json

olafurpg:
The old URL no longer exists.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,4 +1,5 @@
 {
+  "fake": "fake",
   "$schema": "https://raw.githubusercontent.com/sourcegraph/sourcegraph/master/client/shared/src/schema/extension.schema.json",
   "name": "code-intel-extensions",
   "description": "Sourcegraph code intelligence extensions",
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://raw.githubusercontent.com/sourcegraph/sourcegraph/master/shared/src/schema/extension.schema.json",
+  "$schema": "https://raw.githubusercontent.com/sourcegraph/sourcegraph/master/client/shared/src/schema/extension.schema.json",
   "name": "code-intel-extensions",
   "description": "Sourcegraph code intelligence extensions",
   "private": true,
--- a/package.json
+++ b/package.json
@@ -1,5 +1,4 @@
 {
-  "fake": "fake",
   "$schema": "https://raw.githubusercontent.com/sourcegraph/sourcegraph/master/client/shared/src/schema/extension.schema.json",
   "name": "code-intel-extensions",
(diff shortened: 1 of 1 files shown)


efritz (approved):

efritz (changes requested):
hold up, title doesn't match contents

Strum355:
Are there meant to be 7 files changed here? :thinking:

shared/lsif/ranges.ts
```diff
@@ -321,6 +348,19 @@ export async function rangesInRangeWindow(
     )
 }
 
+/** Retrieve all ranges in this document that contain code intelligence. */
+export async function underlinesInDocument(
```
efritz, line 352: We explicitly cache range responses in memory because it's an expensive operation. Could we share the same code path as we use in the defs/hovers paths?

shared/lsif/ranges.ts
```diff
@@ -242,6 +242,33 @@ async function hasRangesQuery(queryGraphQL: QueryGraphQLFn<IntrospectionResponse
     return (await queryGraphQL(introspectionQuery)).__type?.fields.some(field => field.name === 'ranges') || false
 }
 
+const underlinesQuery = gql`
+    query Ranges($repository: String!, $commit: String!, $path: String!, $startLine: Int!, $endLine: Int!) {
```
efritz, line 246: We may want to have a simpler graphql query that doesn't need to aggregate so much data if all we want is all ranges in a document.

shared/activate.ts
```diff
@@ -98,6 +102,34 @@ export async function activateCodeIntel(
     }
 }
 
+export async function activateLsifUnderlines(context: sourcegraph.ExtensionContext): Promise<any> {
```
efritz, line 105: Would be good to talk with the @sourcegraph/web team to see if there is an easy path forward with the extension host so they can control the highlighting instead of changing the cssText of elements directly.

olafurpg:
lol I pushed a WIP commit for an unrelated prototype.

shared/activate.ts
olafurpg: I force pushed this in the middle of a call with @AlicjaSuska to discuss how to move forward with this prototype. This was not at all ready for review yet 😅

shared/lsif/ranges.ts
olafurpg: Yeah, I suspected there's a significantly faster way to get only the ranges. This was just a prototype I was hacking on to demonstrate a proof-of-concept to underline LSIF ranges. I'll make more noise about it another day once it's ready for review :)

efritz (approved):

shared/activate.ts
efritz: 😛

PR #567, opened 2021-02-03
Expand C-family docstring resolution

### LINKED ISSUE #17925: Search-based code intel hovers are broken when definition token and docstring are separated by multiple lines
See for an example https://sourcegraph.com/github.com/CloverHackyColor/CloverBootloader/-/blob/MdeModulePkg/Core/Pei/Hob/Hob.c#L25:1

Generally we try to find the docstring based on searching around the definition token, but the following is also legal, which causes the docstringa nd the definition token to be separated by a larger number of lines + tokens than we expect.

```
<docstring>
<return type>
<visibility modifier>
<actual definition>(<parameter list>) ...
```

Original report from Twitter: https://twitter.com/DrJohnZoid/status/1356598073187733504

efritz:
**Before**:

Screen Shot 2021-02-03 at 2 38 02 PM

**After**:

Screen Shot 2021-02-03 at 2 37 41 PM

**Summary**:

Each supported language in these extensions have a set of properties attached to them (file extensions, what identifier patterns look like, what comments look like, etc). One of these properties is `docstringIgnore`, which will allow us to ignore the lines between where a ctags definition occurs and the comment that is meant to belong to it. Until now, all we've used this for is for Java-style annotations, e.g.

```java
/** docstring **/
@MyBigAnnotation
class Something {
    // ...
}
```

This adds support for the C-family convention of breaking definitions up onto multiple lines:

```cpp
/** docstring **/
RET_VALUE
ActualMethod(...) {
    // ...
}
```

This _could_ have unintended side-effects where comments that do not belong to a ctags definition just happens to be close enough to it, but I think it's an acceptable risk tradeoff (while hand-testing I couldn't get it to do something _worse_ than the previous behavior).

Fixes https://github.com/sourcegraph/sourcegraph/issues/17925.

### DIFF
--- a/shared/language-specs/cpp.ts
+++ b/shared/language-specs/cpp.ts
@@ -48,7 +48,17 @@ export const cppSpec: LanguageSpec = {
         'pc', // Pro-C by Oracle RDBMS
         'pcc', // Pro-C by Oracle RDBMS
     ],
-    commentStyles: [cStyleComment],
+    commentStyles: [
+        {
+            ...cStyleComment,
+            // Ignore identifiers between definition line and doc block. This enables
+            // correctly finding doc strings for function definitions that span multiple
+            // lines (return type and function name separated by a newline). We'll only
+            // enable this for C-family languages as it seems to be the most pervasive
+            // language with this style.
+            ...{ docstringIgnore: /^[\s\w]+/ },
+        },
+    ],
     filterDefinitions,
     lsifSupport: LSIFSupport.Experimental,
 }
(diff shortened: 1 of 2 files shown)


tjdevries:
Side question: How to add a test for something like this (is it simple?)

efritz:
@tjdevries It's difficult to test changes to the language spec properties, so we default back to testing by hand. In order to run the extensions, you can [side load](https://docs.sourcegraph.com/extensions/authoring/local_development) the extension so it can be run with a live SG instance (I used sourcegraph.com here as a user reported an actual line of code, which is always nice).

It would be nice to add integration tests here, but it could be a bit heavy-handed as it requires an instance (and search results from the backend). There are a good amount of tests in the machinery though, so I'm confident we didn't regress on things _outside_ of the C-family heuristics that were changed here.

garobrik (approved):
LGTM

PR #579, opened 2021-03-01
Update extensions reference action item copy

umpox:
As an outcome of the design updates on https://github.com/sourcegraph/sourcegraph/issues/9740

cc: @AlicjaSuska


AlicjaSuska (approved):
Thank you @umpox !

efritz (approved):
FYI these settings only apply to LSP providers and are useless otherwise. We should think about removing them completely.

PR #580, opened 2021-03-03
Add aggregable badges to response

efritz:
Sister PR: https://github.com/sourcegraph/sourcegraph/pull/18843.

**TODO**:

- [x] remove casts to any after we can upgrade sourcegraph type def package


shared/providers.ts
```diff
@@ -286,8 +290,15 @@ export function createReferencesProvider(
...
-            for await (const lsifResult of lsifProvider(textDocument, position, context)) {
-                if (nonEmpty(lsifResult)) {
+            for await (const rawResult of lsifProvider(textDocument, position, context)) {
+                if (nonEmpty(rawResult)) {
+                    const lsifResult = asArray(rawResult).map(
+                        result =>
+                            (({
+                                ...result,
+                                aggregableTags: [{ text: 'semantic', linkURL: HoverAlerts.linkURL }],
+                            } as any) as sourcegraph.Location)
```
Strum355, line 300: are both of these needed? :thinking:

efritz: Yes right now this is v. v. stupid (there's no type intersection so you have to upcast to top before downcasting to sourcegraph.Location).

Once we publish the type def in the sister PR we won't need any casting.

Strum355 (approved):
Pending my comment on the sister PR :heavy_check_mark:

shared/indicators.ts
```diff
@@ -0,0 +1,178 @@
...
+    linkURL,
+    hoverMessage:
+        'It looks like this symbol is defined in another repository that does not have a pre-computed semantic index. Go to definition may be imprecise.',
+}
+
+export const partialDefinitionNoHoverBadge: AggregableBadge = {
+    text: 'partial semantic',
+    linkURL,
+    hoverMessage:
+        'It looks like this symbol is defined in another repository that does not have a pre-computed semantic index. This hover text may be imprecise.',
```
shrouxm, line 42: not sure if the right place for convo about this text is elsewhere, but 
`s/a pre-computed semantic index/semantic indexing/g`?

efritz: This was your text! We can change it to whatever we like:

- It looks like this symbol is defined in another repository that does not have a pre-computed semantic index.
- It looks like this symbol is defined in another repository that does not have semantic indexing.
- It looks like this symbol is defined in another repository that does not have semantic indexing enabled.
- It looks like this symbol is defined in another repository that is not being semantically indexed.

efritz: cc @AlicjaSuska

Strum355 (approved):

PR #588, opened 2021-03-25
Ensure that we remove rejected promises from the cache once resolved

### LINKED ISSUE #19433: Permanent HTTPStatusError after transient error
#### Steps to reproduce:

1. Start enterprise instance
2. Navigate to a blob view (I don't _think_ this code needs a precise index to replicate)
3. Hover over some things and get those sweet, sweet hover messages
4. Restart your instance
5. Hover over things while instance is down, note the console messages and missing results
6. Hover over things after instance comes back up, note continued console messages and missing results

Additional notes:

- If we open the same blob view in a new tab (or refresh) it becomes functional again
- If we navigate away from that document, the new document works. If we navigate back to the old document, it's still broken.
- We see these error messages in the console:

image

#### Expected behavior:

The extension should be able to access the Sourcegraph instance API after a transient network error.

#### Actual behavior:

It isn't.

efritz:
Fixes https://github.com/sourcegraph/sourcegraph/issues/19433.

We were caching promises that failed in the ranges cache, which means that the provider blows up on every subsequent request that requests a position within that range (until it is pushed out of the cache).

This change ensures that if a promise within the ranges result cache is rejected, it will immediately be removed from the cache.

### DIFF
--- a/shared/lsif/ranges.test.ts
+++ b/shared/lsif/ranges.test.ts
@@ -19,9 +19,9 @@ describe('findOverlappingWindows', () => {
 
     it('finds overlapping ranges', async () => {
         const windows = [
-            { startLine: 1, endLine: 3, ranges: Promise.resolve([aggregate1]) },
+            { startLine: 1, endLine: 4, ranges: Promise.resolve([aggregate1]) },
             { startLine: 4, endLine: 6, ranges: Promise.resolve([aggregate2]) },
-            { startLine: 7, endLine: 9, ranges: Promise.resolve([aggregate3]) },
+            { startLine: 6, endLine: 9, ranges: Promise.resolve([aggregate3]) },
         ]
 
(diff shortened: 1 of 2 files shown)


Strum355:
Im not sure how this works 🤔 quick tldr of why these changes specifically?

efritz:
@Strum355 

So we have a map from `key => Promise<T>`. When the network goes down (or we get a 502 during a rolling deploy), any network requests will have put a rejected promise value in the cache. When the network comes back (or the frontend becomes healthy again), we might pull back a rejected promise from the cache and re-throw the error without retrying.

This change makes it so that we put a promise in the cache that later fails, we'll remove it from the cache so we can retry.

We _could_ also just await the ranges request to see if it fails and not stash it if it does. However, we want to put the promise and not the result in the cache because many users may ask for the same key before the promise can resolve. This reduces n ranges request (which are more expensive than definition and hover requests) to a single API request.

tjkandala (approved):

PR #589, opened 2021-03-25
Do not block on telemetry

efritz:
This makes the `emitOnce` function something that will fire-and-forget the telemetry requests so that a long write to the event logs table or a network failure does not blow up the user experience.


Strum355 (approved):
Was wondering why the awaits when they were introduced, thats on me for not bringing that up ✋

efritz:
@Strum355 Makes sense since it _is_ a network request and that's ... generally what you do here.

PR #591, opened 2021-03-25
Add commit to manifest

efritz:
In order to help customers with a private extension registry triage, we need to know what version they have synced. Right now we only have a date that they synced, from which it is a bit clunky to find a commit.

This adds the commit directly to the manifest so we can ask them for the exact version of code they are running.

### DIFF
--- a/dev/scripts/generate.ts
+++ b/dev/scripts/generate.ts
@@ -3,6 +3,10 @@ import * as fs from 'mz/fs'
 import * as path from 'path'
 import { LanguageSpec } from '../../shared/language-specs/spec'
 import { findLanguageSpecs } from './args'
+import { exec as _exec } from 'child_process'
+import { promisify } from 'util'
+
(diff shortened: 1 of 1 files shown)


olafurpg (approved):
LGTM 🚀

felixfbecker:
This is great! Could we add this to the [`src ext publish` command](https://github.com/sourcegraph/src-cli/blob/main/cmd/src/extensions_publish.go) so it works for every extension? That's exactly what `npm publish` does, it adds a `gitHead` field with the commit hash in the containing repo at the time of publish. This could be useful for other extensions too

efritz:
@felixfbecker Should this be done _instead_ of the current change? I'd be more than happy to implement it that way.

efritz:
Will likely close in favor of https://github.com/sourcegraph/src-cli/pull/500.

felixfbecker:
Yeah if that would solve your needs

PR #608, opened 2021-04-23
Add `-` to powershell identifiers

efritz:


Strum355 (approved):
Is this really the entire fix lmao

shared/language-specs/languages.ts
```diff
@@ -202,7 +202,7 @@ const powershellSpec: LanguageSpec = {
     languageID: 'powershell',
     stylized: 'PowerShell',
     fileExts: ['ps1', 'psd1', 'psm1'],
-    identCharPattern: createIdentCharPattern('?'),
+    identCharPattern: createIdentCharPattern('-?'),
```
efritz, line 205:
```suggestion
    identCharPattern: createIdentCharPattern('?-'),
```

efritz:
@Strum355 **No**, I just noticed that this is wrong too and it's a common identifier in functions.

keegancsmith:
Don't know powershell at all, so don't know if this is correct but assume it is :)

PR #615, opened 2021-05-04
Add setting and toggle to omit supplementing reference results with search-based results

efritz:
@olafurpg: Looking (in particular) for better names for settings and text around _Show/Hide search results_


extensions/typescript/package.json
```diff
@@ -73,6 +77,22 @@
         "actionItem": {
           "label": "${config.typescript.showExternalReferences && \"Hide other repositories\" || \"Show other repositories\"}"
         }
+      },
+      {
+        "id": "supplementReferencesWithSearchResults.toggle",
+        "command": "updateConfiguration",
+        "title": "${!!config.codeIntel.supplementReferencesWithSearchResults && \"Hide search results\" || \"Show search results\"}",
```
olafurpg, line 84: I think it would be good to consistently stick with "search-based"
```suggestion
        "title": "${!!config.codeIntel.supplementReferencesWithSearchResults && \"Hide search-based results\" || \"Show search-based results\"}",
```

olafurpg (approved):
LGTM 👍 

Do I understand correctly that this also adds buttons to enable/disable this setting in the UI?

I think `supplementReferencesWithSearchResults` is an OK name. I tried to come up with alternatives but didn't find anything I like better.

olafurpg:
Thank you for commandeering this! 🙏 I was partly demotivated by adding only a hidden configuration option. It's fantastic if we can make this feature visible in the UI

extensions/typescript/package.json
efritz: I think we need something better than either of these. The text should actually read "stop showing search-based results if there are precise ones" but I don't know a very clear and succinct way to say that as a text link.

Strum355:
First round of feedback on trying this locally as a _user_: 
- the default was "search based results disabled" yet it showed search based results. I had to toggle off and on again to get it to work.
- While current position (repo/file-in-repo if grouping) is preserved, the whole panel being wiped and into "loading" state is slightly off-putting but really just a minor nit and probably a lot more work to avoid
Otherwise, all seems good :)

efritz:
@Strum355 Will check default value to make sure it matches. As for the second point, the extensions API doesn't give us enough control over this (the @sourcegraph/extensibility has confirmed that this is the current workaround).

We need to be able to re-invoke the references provider when this setting changes, but if we've already returned we can't do that from inside the extension. The other option is to never return from the provider, but it will appear as if the panel is indefinitely loading.

Re-iterating here in case that bumps any work up in priority to support this use case.

extensions/go/package.json
```diff
@@ -50,6 +50,10 @@
         {
           "action": "externalReferences.toggle",
           "when": "panel.activeView.id == 'references'"
+        },
+        {
+          "action": "mixPreciseAndSearchBasedReferences.toggle",
```
olafurpg, line 55: Is the name "precise" used anywhere in the product? I think it would be better to use "semantic" for consistency with the labels in the UI
```suggestion
          "action": "mixSemanticAndSearchBasedReferences.toggle",
```

efritz: The only place semantic is used is in the hover, all of our docs use precise. Maybe we should wait to merge until after we conduct a user study and nail down naming.

efritz: Precise has scored better over semantic in user-testing. This should also be a fairly easy thing to change in the future without leaving too much junk around (this is a configuration setting controlled by a UI button and shouldn't generally be changed by hand in the user settings page).

PR #621, opened 2021-05-11
bundle repository ID in telemetry event to not depend on url in backend

Strum355:
As a caching mechanism already existed at a different level, I've moved it lower down to the `API` class so it can be cached for both the search provider and the general code-intel feature providers.

This is part of the work for private code on dotcom, specifically https://github.com/sourcegraph/sourcegraph/issues/20628. The reliance on the `argument` column is still necessary, further discussion can be found on Slack

### DIFF
--- a/shared/telemetry.ts
+++ b/shared/telemetry.ts
@@ -7,13 +7,24 @@ import * as sourcegraph from 'sourcegraph'
  */
 export class TelemetryEmitter {
     private languageID: string
+    private repoID: number
     private started: number
     private enabled: boolean
     private emitted = new Set<string>()
 
-    constructor(languageID: string, enabled = true) {
+    /**
+     * Creates a new telemetry emitter object for a given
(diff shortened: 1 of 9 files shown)


shared/telemetry.ts
```diff
@@ -7,13 +8,21 @@ import * as sourcegraph from 'sourcegraph'
...
 export class TelemetryEmitter {
     private languageID: string
+    private repoID: string
     private started: number
     private enabled: boolean
     private emitted = new Set<string>()
 
-    constructor(languageID: string, enabled = true) {
+    /**
+     *
```
efritz, line 17: Stupid nit, could you put in a dumb sentence here? (Also the repoID docstring isn't complete).

efritz (approved):
Assumed you side-loaded the extension and made sure it still ran?

Strum355:
Yup indeed

PR #680, opened 2021-09-08
Revert LSP deprecation

efritz:
Seeing `Provider errored: RangeError: Invalid array length` on Cloud after this merge.


efritz:
Found fix in https://github.com/sourcegraph/code-intel-extensions/pull/681.

PR #681, opened 2021-09-08
Fix illegal spread

efritz:
Not sure why we get a `RangeError` from the previous code.

```
RangeError: Invalid array length
    at __spreadArrays (http://localhost:1234/extension.js:41846...
```


Strum355 (approved):
javascript moment

PR #682, opened 2021-09-08
Fix illegal spread (part 2)

efritz:


.eslintrc.js
```diff
@@ -8,5 +8,9 @@ module.exports = {
     project: ['template/tsconfig.json', 'dev/scripts/tsconfig.json'],
   },
   ignorePatterns: ['generated-*', '**/dist', 'samples'],
-  rules: {},
+  rules: {
+    // [...a] breaks when Array.from(a) works
+    // when a is not an array but an iterable
+    "unicorn/prefer-spread": 0
```
Strum355, line 14: Should we also remove the inline eslint comment from the last PR?

Strum355 (approved):

.eslintrc.js
efritz: See template/src/search/tokens.ts

Numbers88s:
Don't mind me @Numbers88s [Zapier related]

PR #686, opened 2021-09-23
Add Apex support

chrismwendt:


template/src/language-specs/languages.ts
```diff
@@ -20,6 +20,13 @@ import { typescriptSpec } from './typescript'
 
 // Please keep the specs below in lexicographic order.
 
+const apexSpec: LanguageSpec = {
+    languageID: 'apex',
+    stylized: 'Apex',
+    fileExts: ['apex', 'cls'],
```
tjdevries, line 26: I thing `trigger` is also a viable option?

chrismwendt:
Got approval in Slack ✅

PR #687, opened 2021-09-24
Don't put \b in the query unless the token begins/ends with \w

chrismwendt:
Bug report https://github.com/sourcegraph/sourcegraph/issues/25322

Previously, we'd search for `\bfoo!\b` in Ruby, which matches nothing because `!\b` seldom (if ever) matches anything.

This changes our queries so that we'll search for `\bfoo!` in Ruby.

### DIFF
--- a/template/src/search/queries.ts
+++ b/template/src/search/queries.ts
@@ -1,6 +1,6 @@
-import { escapeRegExp } from 'lodash'
 import { extname } from 'path'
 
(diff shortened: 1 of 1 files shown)


efritz (approved):
DAMN IT I WAS JUST DOING THIS

chrismwendt:
2fast 😛 I put eyes emoji on the Slack message, so I basically called dibs.

PR #688, opened 2021-09-24
Increase resolution of Apex icon

chrismwendt:


efritz (approved):
I'll trust the PR title

efritz:
Yes, proof it is indeed bigger.
Screen Shot 2021-09-24 at 3 40 39 PM

chrismwendt:
CleanShot 2021-09-24 at 14 40 52@2x

PR #701, opened 2021-10-18
Stencil

efritz:
Here's a bit of hacking I did to start reducing network chatter. This uses a cached "stencil" (a set of ranges with hover data) for each text document and returns `undefined` quickly if the given position is not in the stencil. This adds a blocking call to the _first_ `DefinitionAndHover` request to every document.

Remaining tasks:
- [x] Add limit to stencil cache
- [x] Update tests (if failing), add tests (if uncovered)
- [x] Update combined providers to *not* make search requests for documents with LSIF data (must be completed in order for network chatter to be reduced)
- [x] Check for the `stencil` GraphQL endpoint for old Sourcegraph instances


chrismwendt:
I added stencil checks in refs and doc highlights. Does that complete the task "Update combined providers to not make search requests for documents with LSIF data"?

chrismwendt (approved):
Sweeeet ✨

PR #724, opened 2021-11-15
Only show "Find implementations" for languages that support it

chrismwendt:
Forgot to do hide the button for languages that we know don't support implementations in #683


cesrjimenez (approved):
LGTM 🚀

PR #726, opened 2021-11-19
spike: Reduce network chattiness

efritz:
Ideas:

- [ ] When we catch _no LSIF data_ errors, we should mark that text document as not having LSIF data for a short amount of time and short-circuit the LSIF providers. Right now we'll hit the LSIF resolvers on all hover actions just to check that there are no uploads to service it. We can skip re-hitting this endpoint to get the same error, but re-check every 30s or so in case an upload comes in asynchronously.
- [ ] When we have LSIF data for a text document but do _not_ have a hover or definition for a particular position, we shouldn't fall back to search. This gives us hovers on things like _error_ in Go that are not linked to the stdlib which just jumps to a random line. Feels broken. (We should still mix precise and search-based _references_ though, and there are already explicit user flags to control this behavior).

### DIFF
--- a/template/src/lsif/api.ts
+++ b/template/src/lsif/api.ts
@@ -22,9 +22,14 @@ export async function queryLSIF<P extends { query: string; uri: string }, R>(
         ...rest
     }: P,
     queryGraphQL: QueryGraphQLFn<GenericLSIFResponse<R>>
-): Promise<R | null> {
+): Promise<R> {
     const { repo, commit, path } = parseGitURI(new URL(uri))
     const queryArguments = { repository: repo, commit, path, ...rest }
     const data = await queryGraphQL(query, queryArguments)
-    return data.repository?.commit?.blob?.lsif || null
+    const lsifData = data.repository?.commit?.blob?.lsif
+    if (!lsifData) {
+        throw new Error('No LSIF data')
+    }
+
+    return lsifData
 }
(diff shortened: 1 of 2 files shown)


Strum355 (approved):
theory: good :heavy_check_mark: 
code: maybe :slightly_smiling_face:

efritz:
@Strum355 Let's just convert this back to draft since it was just a way to play with the PR description 😛

chrismwendt:
What @Strum355 said 😅 Another possibility: mark the document as not having LSIF until it gets reopened. That might be easier to implement and understand the behavior because a timer isn't involved.

mrnugget (commented):
I like the proposed behaviour. _Ideally_ we'd have a way to let the user know about these two states

1. "text document as not having LSIF data"
2. "we have LSIF data for a text document but do not have a hover or definition for a particular position"

the latter one is what I ran into in that rust repo and it was super confusing.

(Can't comment on the code but will now try my hardest to see how to run this code locally and play around with it)

mrnugget:
Okay, I read through the code again, this time knowing kinda how the extension works. I think this approach makes sense. Of course things would be easier, I guess, if we didn't have to follow that architecture.

Clarifying question:

> When we have LSIF data for a text document but do not have a hover or definition for a particular position, we shouldn't fall back to search. 

So that means: in the `definitionAndHover` provider we never fall back to search? I don't understand when it's the case that we "have LSIF DATA for a text document but do not have a hover [...]"?

efritz:
@mrnugget 

This can happen if `break` happens to return search results, for example. I've seen it happen with ranges that weren't actually symbols that should have data, just keywords like null or primitive types.

mrnugget:
Ah, I see. So we have indexed the document, someone hovers over `break`, we check the LSIF provider for `definitionAndHover`, don't get results and fallback to search.

Question then is: how do we accurately determine whether "we have LSIF data for a text document" if the user, for example, hovers over `break` first thing?

PR #743, opened 2022-01-20
Fix duplicate find-implementations button

chrismwendt:
Fixes https://github.com/sourcegraph/sourcegraph/issues/28973

@tjkandala Is this what you had in mind when you said ["Could they check that the current language also matches the extension's language?"](https://github.com/sourcegraph/sourcegraph/pull/29853#issuecomment-1016606604)?

### DIFF
--- a/template/package.json
+++ b/template/package.json
@@ -53,7 +53,7 @@
       "hover": [
         {
           "action": "findImplementations",
-          "when": "get(context, `implementations_${resource.language}`) && (goToDefinition.showLoading || goToDefinition.url || goToDefinition.error)"
+          "when": "resource.language == 'LANGID' && get(context, `implementations_${resource.language}`) && (goToDefinition.showLoading || goToDefinition.url || goToDefinition.error)"
         }
       ]
     },


philipp-spiess:
Yeah I think this is the idea: If two different hover find-implementations actions are loaded but we hide those with the language not active, we shouldn't be seeing the issue. cc @tjkandala 

I tried to repro the issue with your changes following the [extension dev guide](https://github.com/sourcegraph/code-intel-extensions#development) but it seems like Find implementation isn't showing up at all now:

Screenshot 2022-01-20 at 13 05 18

Maybe this has to do with the extension having the name `template` instead of being language specific. Even if this would, we would likely not be able to repro the issue unless we install two language-specific extensions:
 
Screenshot 2022-01-20 at 13 05 35

If you have some ideas on how I could verify that this fixes the issue, I’m happy to dig in!

template/package.json
```diff
@@ -53,7 +53,7 @@
       "hover": [
         {
           "action": "findImplementations",
-          "when": "get(context, `implementations_${resource.language}`) && (goToDefinition.showLoading || goToDefinition.url || goToDefinition.error)"
+          "when": "resource.language == 'LANGID' && get(context, `implementations_${resource.language}`) && (goToDefinition.showLoading || goToDefinition.url || goToDefinition.error)"
```
philipp-spiess, line 56: I wonder if we also want to include this condition with the `menus.panel/toolbar` item above? Basically anything that defines a when condition here will be exported as _any_ language extension so there might be other collisions.

mrnugget:
@philipp-spiess I ran into the same problem just now (and your comment popped up while my comment was sitting here in the text box). Locally I changed the `LANGID` in the `package.json` to `go` to test this.

I get the button then but I can also reproduce the original issue. Not sure if that's due to the "template" language being active on all languages and the Rust extension also still being active?

screenshot_2022-01-20_13 14 20@2x

mrnugget:
If I disable the "Rust" extension on Sourcegraph.com ("Go" was already disabled) and sideload the extension from localhost with `LANGID` in the check replaced with `go` it seems to be fixed? But again: not sure whether that's due to funky local setup.

philipp-spiess:
> @philipp-spiess I ran into the same problem just now (and your comment popped up while my comment was sitting here in the text box)

Ahaha great minds 🙂 

> I get the button then but I can also reproduce the original issue. Not sure if that's due to the "template" language being active on all languages and the Rust extension also still being active?

So to repro, you jumped from Go -> Rust or from Rust -> Go?

I believe to test that this is fixed we need to switch __from__ the sideloaded extension to a __default__ one (or sideload both). So maybe you can try to set the `LANGID` to `rust` instead and only disable the sideloaded extension?

mrnugget:
To repro I followed the exact steps here https://github.com/sourcegraph/sourcegraph/issues/28973#issuecomment-1014695100 after disabling with extensions ("Go" and "Rust") locally and side-loading `template` (which I think acts as if it's available for every language?)

philipp-spiess:
> To repro I followed the exact steps here [sourcegraph/sourcegraph#28973 (comment)](https://github.com/sourcegraph/sourcegraph/issues/28973#issuecomment-1014695100) after disabling with extensions ("Go" and "Rust") locally and side-loading `template` (which I think acts as if it's available for every language?)

I think the problem is that with this approach, we only register __one extension__ but for the issue to reproduce we need two different extensions with the same exported menu items. The console would need to show `Activating Sourcegraph extension: template` twice.

mrnugget:
Ahh, I see. Maybe we can generate the actual extensions locally and serve them?

template/package.json
tjkandala: I think so! Then we could get rid of the band-aid action deduplication in the `<Panel>` component

tjkandala:
>@tjkandala Is this what you had in mind when you said "Could they check that the current language also matches the extension's language?"?

Yes, fix looks good 👍

> Ahh, I see. Maybe we can generate the actual extensions locally and serve them?

Probably the best (only?) way to confirm that it works!

chrismwendt:
Local repro steps are tricky. We can't use the usual `yarn --cwd template run serve` because it doesn't replace `'LANGID'` in `template/package.json`, so `resource.language == 'LANGID'` is always false, which is why you don't see the "Find implementations" button. I'm going to change the local dev scripts so we don't hit this snag in the future.

Here's what I did

1. Generate and serve the `rust` extension:

```
yarn run generate --languages=rust
cd generated-rust
yarn run serve
```

2. Sideload it
3. Set `{ "extensions": { "sourcegraph/rust": false } }` in settings https://sourcegraph.com/users/chris/settings
4. Visit a Rust file
5. Navigate through Sourcegraph to a Go file (you can't refresh the page, you have to stay in the SPA)
6. You should see 1 "Find implementations" button

Prior to this video I deleted `resource.language == 'rust'`, then during the video I restore it and show that it prevents showing duplicate buttons:

https://user-images.githubusercontent.com/1387653/150415462-f95efc80-ab50-492e-b87c-b08406311fe1.mp4

tjkandala (approved):
Nice!

PR #753, opened 2022-02-07
Update to open alert on new window

### LINKED ISSUE #30518: Re-direct users in a new window when clicking on precise code intel banner in popover
#### Feature request description

image (2)

Currently, clicking on the "learn more about precise code intel" button link within the code intel popover redirects users to the user docs, but in the same window. This takes users out of context of the code that they're looking at, and it potentially impacts our retention for first-time Sourcegraph users (or early users) when learning more about our product. 

The request is to open the precise code intel user docs in a **new window**, so that users can maintain their file view and context, while being able to learn more about precise code intel.

#### Additional context

Raised in [internal feedback](https://sourcegraph.slack.com/archives/CHXHX7XAS/p1643636773057509).

erzhtor:
Closes https://github.com/sourcegraph/sourcegraph/issues/30518.

## How to test
- Run `yarn run generate --languages=typescript && yarn --cwd generated-typescript run serve`
- Go to sourcegraph.com
- Remove `WebHoverOverlay.dismissedAlerts` from localStorage
- Add `"sourcegraph/typescript": false` to `user settings -> extensions`
- Open any [TypeScript file](https://sourcegraph.com/github.com/sourcegraph/code-intel-extensions/-/blob/template/src/indicators.ts?L64) and hover over a token
- Make sure "Learn more ..." open on a new tab
- Don't forget to remove `"sourcegraph/typescript": false`

### DIFF
--- a/template/src/indicators.ts
+++ b/template/src/indicators.ts
@@ -64,7 +64,7 @@ const makeAlert = ({
         iconKind: 'info',
         summary: {
             kind: sourcegraph.MarkupKind.Markdown,
-            value: `${message}<br /> [Learn more about precise code intelligence](${linkURL})`,
+            value: `${message}<br /> <a href="${linkURL})" target="_blank">Learn more about precise code intelligence</a>`,
         },
         ...legacyFields,
     }


vdavid:
We also thought about making this more generic (to automatically open all external links in a new tab), but then we went with this single fix.

mrnugget (approved):
Thanks a lot! (Not sure if the `.vscode/settings.json` should be in here? But doesn't hurt, i guess.)

erzhtor:
> Thanks a lot! (Not sure if the `.vscode/settings.json` should be in here? But doesn't hurt, i guess.)

Sorry, fixed! Thanks for catching it! It should not be committed; my local extension auto-added it.

mrnugget:
I think you can go ahead and merge, too, @erzhtor

PR #756, opened 2022-02-08
Configure renovate to stop tagging code intel for reviews

olafurpg:
These notifications create a lot of noise with the GitHub Slack
integration.

PR #757, opened 2022-02-13
Configure renovate to stop tagging code intel for reviews

varungandhi-src:
These notifications create a lot of noise with the GitHub Slack
integration.

This is the same as #756 but with the commit in the main repo instead of
in a fork, since CI testing doesn't run for PRs from forks.


varungandhi-src:
Merging with an explicit approval since the same PR was already approved earlier.

PR #790, opened 2022-05-06
Add request for non-local variable definition

chrismwendt:
Part of https://github.com/sourcegraph/sourcegraph/issues/35031

Depends on https://github.com/sourcegraph/sourcegraph/pull/35082

This requests a non-local variable definition if a local one could not be found.

**Reviewers: ignore the GraphQL changes - they're synced from sourcegraph/sourcegraph**

### DIFF
--- a/schema/schema.graphql
+++ b/schema/schema.graphql
@@ -2280,6 +2280,7 @@ enum ExternalServiceKind {
     AWSCODECOMMIT
     BITBUCKETCLOUD
     BITBUCKETSERVER
+    GERRIT
     GITHUB
     GITLAB
     GITOLITE
@@ -2290,6 +2291,7 @@ enum ExternalServiceKind {
     PAGURE
     PERFORCE
     PHABRICATOR
+    PYTHONPACKAGES
 }
 
 """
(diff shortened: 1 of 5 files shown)


schema/schema.graphql
```diff
@@ -2290,6 +2291,7 @@ enum ExternalServiceKind {
     PAGURE
     PERFORCE
     PHABRICATOR
+    PYTHONPACKAGES
```
varungandhi-src, line 2294: Q: How is this change + the change adding GERRIT above related to the rest of this patch? Did those changes sneak in through rebasing?

schema/schema.graphql
```diff
@@ -3768,6 +3770,15 @@ type GitCommit implements Node {
...
     """
+    The Git tree or blob in this commit at the given path.
+    """
+    path(
+        """
+        The path of the tree or blob.
+        """
+        path: String = ""
+    ): GitTreeOrBlob
+    """
```
varungandhi-src, line 3781: [Warning: GraphQL noob question] I'm a little confused... There is already a `GitTree` below and a `GitBlob` as well. Could you clarify why this field is needed?

schema/schema.graphql
chrismwendt: This GraphQL file hasn't been updated in a while, so syncing it pulled in a lot of unrelated changes.

schema/schema.graphql
chrismwendt: (this is an unrelated change that got synced)

`GitTreeOrBlob` appears to be an alias for `GitTree | GitBlob`.

cesrjimenez (approved):
LGTM! 🚀

PR #791, opened 2022-05-09
ci: Add GraphQL schema syncing job

varungandhi-src:
Stops the schema from becoming out-of-date over time.

### Test Plan

n/a

### DIFF
--- a/.github/workflows/sync-graphql.yml
+++ b/.github/workflows/sync-graphql.yml
@@ -1,0 +1,24 @@
+name: sync-graphql
+on:
+  schedule:
+    - cron: '0 8 * * *'
+
+jobs:
+  sync-graphql:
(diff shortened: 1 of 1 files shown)


varungandhi-src:
@bobheadxi I've tagged you in case you have better ideas on how to accomplish this. 😅

bobheadxi:
The alternative would be to have `sourcegraph/sourcegraph` trigger an update, e.g. what @davejrt did for Zoekt: https://github.com/sourcegraph/zoekt/blob/master/.github/workflows/pr.yml

If this works it works though :)

varungandhi-src:
Oh yeah, that seems like a better approach, because that does it on push, instead of a periodic basis. It also uses auto-merging instead of requiring manual approval.

PR #797, opened 2022-05-23
Implement document highlight fallback for search-based intel

olafurpg:
Fixes https://github.com/sourcegraph/sourcegraph/issues/35843

### DIFF
--- a/template/src/search/providers.ts
+++ b/template/src/search/providers.ts
@@ -292,7 +292,38 @@ export function createProviders(
         if (squirrelDocumentHighlights) {
             return squirrelDocumentHighlights
         }
-        return null
+        const contentAndToken = await getContentAndToken(textDocument, position)
+        if (!contentAndToken) {
+            return null
+        }
+        const { text, searchToken } = contentAndToken
+        const result: sourcegraph.DocumentHighlight[] = []
+        let line = 0
+        let lastLineIndex = 0
+        const tokenPattern = new RegExp('\\b' + searchToken + '\\b', 'g')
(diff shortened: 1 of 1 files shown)


template/src/search/providers.ts
```diff
@@ -292,7 +292,38 @@ export function createProviders(
...
+        if (!contentAndToken) {
+            return null
+        }
+        const { text, searchToken } = contentAndToken
+        const result: sourcegraph.DocumentHighlight[] = []
+        let line = 0
+        let lastLineIndex = 0
+        const tokenPattern = new RegExp('\\b' + searchToken + '\\b', 'g')
+        const matches = [...text.matchAll(tokenPattern)]
+        // TODO: why is matches empty?
```
olafurpg, line 305: I don't understand why `matches` is always empty here. Any advice on what I'm doing wrong? When I test the same regexp manually then it works as expected, it's only empty when running via webworkers 🤔

efritz: Do you know the values of text and pattern here at runtime?

olafurpg: Yes. I even used "copy object" on the actual logged values and `matchAll()` worked as expected against that object in the JS console. The `text` value has the full multi-line string value of the document and the pattern is correct too.

chrismwendt (approved):
Sweet! 🚀

efritz:
@olafurpg Do we want to close and/or transfer the code in this PR to sg/sg?

olafurpg:
Closing since we should implement this feature in sg/sg

PR #798, opened 2022-05-24
Enable "Find implementations" for TypeScript

olafurpg:
See https://github.com/sourcegraph/scip-typescript/pull/136


varungandhi-src (approved):
Also, we should make a matching change in sg/sg's docs.

olafurpg:
Done https://github.com/sourcegraph/sourcegraph/pull/35966

PR #817, opened 2022-07-15
Fix off-by-one in range inclusion

### LINKED ISSUE #33328: code-intel: Whitespace characters are sometimes hoverable
- Visit https://sourcegraph.com/github.com/sourcegraph/lsif-java@5d7ccd55d13282980b843c94dbe5ce88786e2cc3/-/blob/lsif-semanticdb/src/main/java/com/sourcegraph/lsif_semanticdb/SignatureFormatter.java?L288

CleanShot 2022-04-01 at 17 01 38@2x

Not sure exactly what's going on here, but I'm guessing it's an off-by-1 issue. I noticed this right after deploying [local code intel with tree-sitter](https://github.com/sourcegraph/sourcegraph/pull/32122), so that's the first place I'd look.

chrismwendt:
Fixes https://github.com/sourcegraph/sourcegraph/issues/33328

### DIFF
--- a/template/src/util/api.ts
+++ b/template/src/util/api.ts
@@ -682,7 +682,7 @@ const isInRange = (position: sourcegraph.Position, range: Range): boolean => {
     if (position.character < range.column) {
         return false
     }
-    if (position.character > range.column + range.length) {
+    if (position.character >= range.column + range.length) {
         return false
     }
     return true


efritz (approved):
pft yeah probably

